PR #5, opened 2016-04-07
Test with tox + pytest

adamchainz:
This allows the tests to be run on multiple versions of python.

Depends on #4 .


adamchainz:
This currently doesn't work as `realistic_points` doesn't exist

```
________________________________________________________ ERROR collecting test/test_it.py _________________________________________________________
test/test_it.py:6: in <module>
    from realistic_points import realistic_points
E   ImportError: No module named 'realistic_points'
```

adamchainz:
You realize this is broken?

MrMinimal64:
Not for now. I forgot to include this file with the package. I am just fixing it.

PR #13, opened 2016-04-08
setup.py doesn't use download_url, use universal wheels

adamchainz:
* Upload the package to PyPI rather than leaving it on Github so it's easier to cache in downstream PyPI caches
* Use universal wheels so the package can be made into a single wheel that works on Python 2 and 3

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
 [metadata]
 description-file = README.rst
+
+[wheel]
+universal = 1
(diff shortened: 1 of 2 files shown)


MrMinimal64:
the package is on PyPI already:
https://pypi.python.org/pypi/timezonefinder/

or what do you mean?

adamchainz:
`download_url` allows a package to exist on PyPI and get downloaded from elsewhere, just because you're uploading to PyPI doesn't mean people get it from there.

Also you should run `python sdist bdist_wheel upload` as well to generate both the package types and upload, you'll see both the `.tar.gz` and the `.whl` being created, with newer `pip` versions preferring the wheels. They're smaller and more compatible, see e.g. my project https://pypi.python.org/pypi/django-mysql - wheel is 1/3 the size

PR #14, opened 2016-04-08
Fix packaging - include README.rst so setup.py works

adamchainz:
You can test your setup.py is working with `tox`, which does a full `setup.py install` in the test virtualenvs


adamchainz:
@MrMinimal64 bump

PR #25, opened 2016-04-18
Tests - add runtests.py that runs linter programs and py.test

adamchainz:
This is a script adapted from my other projects. It runs several useful linter programs to keep source consistent and detect errors:

* `flake8` - keeps code in line with PEP8 plus detects potential errors with pyflakes
* `2to3` - a built-in Python tool that can help detect forwards-compatibility fixes for Python 2. Not all are activated since some subtly change the semantics.
* `isort` - keeps imports in order, and is configured to add the `__future__` imports for forwards compatibility in Python 2.7
* `setup.py check` - checks your `setup.py` is correctly configured and that your `long_description` (README.rst) will render correctly on PyPI

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,10 @@
+docutils==0.12
+flake8==2.5.4
+isort==4.2.5
 numpy==1.11.0
+pep8==1.7.0
 py==1.4.31
+pyflakes==1.1.0
+Pygments==2.1.3
 pytest==2.9.1
 tzwhere==2.2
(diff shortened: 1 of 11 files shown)


MrMinimal64:
ok, some changes are reasonable, but do you really think those few small tests justify having that much more dependencies?

adamchainz:
These dependencies don't affect your end users, only the development environment. The more linters the better, the dependencies are fine to achieve this - they're all very stable and change rarely because they're widely used.

Issue #40, opened 2016-07-06
Using ahead of time compilation


MrMinimal64:
Make use of the pycc feature of Numba:
http://numba.pydata.org/numba-doc/latest/user/pycc.html

It would be ideal if the helper functions were compiled on installation on the user's system once (when Numba is installed). This would probably speed up the algorithms.

I encountered the issue that on my system the produced (compiled) code is for the wrong target architecture and hence cannot be run. 

I would be glad if someone else could tell me if they are encountering the same issues.

A draft for how this could work is already included in the latest version of timezonefinder.
Procedure would be:
- uncomment the compilation code in helpers_numba.py
- uncomment the import code in timezonefinder.py
- use timezonefinder (then you should get notifications if it worked or not)

MrMinimal64:
I do not consider this to be important any more.

PR #44, opened 2017-01-31
Fix numba for py27

synapticarbors:
I saw the issue reported in https://github.com/MrMinimal64/timezonefinder/issues/43 when I was reviewing the conda-forge recipe, so I decided to go ahead and fix it. I'm not sure of the origin, but it was reproducible. I just changed the jit function signatures from a string to using the object representations of the numba types. Typically, when I use numba, I don't even bother providing a type signature and just let numba figure it out since I haven't seen any really performance improvements in adding them. 

I tested this under python 2.7 and 3.5 and running pytest from the top level directory seems to result in all tests passing. 

cc/ @snowman2

### DIFF
--- a/test/test_it.py
+++ b/test/test_it.py
@@ -4,9 +4,8 @@ import random
 import unittest
 from datetime import datetime
 
-from tzwhere.tzwhere import tzwhere
-
 from timezonefinder.timezonefinder import TimezoneFinder
+from tzwhere.tzwhere import tzwhere
 
 # number of points to test (in each test, realistic and random ones)
 N = 100
--- a/test/test_it.py
+++ b/test/test_it.py
@@ -4,8 +4,9 @@ import random
 import unittest
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


MrMinimal64:
Great, thanks! I provided the signature, because later I want to use ahead of time compilation, but I didn't think of using the object representation. At the moment however I could not get that to work. @synapticarbors are you by any chance an expert in Numba?

synapticarbors:
I would call myself an experienced user of Numba, but I'm not an expert about Numba internals. I haven't used the ahead-of-time compilation feature before though.

MrMinimal64:
ok. I would be glad if you could try it and tell me in case you encounter the same issues.
As in issue  #40:
> Procedure would be:
> - uncomment the compilation code in helpers_numba.py
> - run helpers_numba.py
> - uncomment the import code in timezonefinder.py
> - use timezonefinder (then you should get notifications if it worked or not)

PR #190, opened 2023-03-26
parallel data conversion: shortcut computation 

jannikmi:


jannikmi:
FIXME: the polygon lists are empty

Issue #195, opened 2023-03-28
CI/CD: Node.js 12 actions are deprecated


jannikmi:
test steps
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, pre-commit/action@v2.0.3, actions/setup-python@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

Mourishitz:
Hello @jannikmi!!

I have just submited a Pull Request with the node version changes, you can check it here https://github.com/jannikmi/timezonefinder/pull/200

Hope this solves the problem, and if not, I'd be pleasured to hear and contribute to your project

PR #200, opened 2023-05-08
updating node-version for build.yml

Mourishitz:
Making changes for the https://github.com/jannikmi/timezonefinder/issues/195 issue

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
       TOXENV: ${{ matrix.tox-env }}
       TOX_SKIP_MISSING_INTERPRETERS: False
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           # Use Node.js 16.x
           node-version: "16.x"
@@ -56,7 +56,7 @@ jobs:
     needs: test
     if: endsWith(github.ref, '/master')
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
 
       - name: Set up Python
         uses: actions/setup-python@v2
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,8 +29,6 @@ jobs:
 
       - name: Run pre-commit hook
         uses: pre-commit/action@v3.0.0
-        with:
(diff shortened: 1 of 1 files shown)


Mourishitz:
Done! I had to force push in order to reset HEAD~ of a wrong commit I made on my local repository, let me know if any update is required, I look forward on contributing to the project as soon as possible!

jannikmi:
Strange, but it did not work. the warning is still showing. could you please look into how other repositories have fixed this (look at latest github action workflows)

Mourishitz:
Sure, I'll do it and come back with an answer (and a commit if necessary)

Mourishitz:
It seems that the actions packages are not all available in version 3 already, I looked trough the latest versions and fixed with the best available version for each one.

I used as my main reference this documentation: https://docs.github.com/en/actions/language-and-framework-guides/using-nodejs-with-github-actions#using-a-specific-nodejs-version
Also, I checked actions versions using these:
- actions/checkout: https://github.com/actions/checkout
- pre-commit/action: https://github.com/pre-commit/action
- actions/setup-python: https://github.com/actions/setup-python

Mourishitz:
Oh, I just saw that, I'll try to run it locally on my fork to find the best ones and find more documentations about the versions

Mourishitz:
I have just refactored some changes, unfortunately I couldn't test it on my local repository due to repository secrets, hope this one do the work for us!

jannikmi:
Thanks. Seems like it worked!

there is a new error:

`Unexpected input(s) 'node-version', `

could you please remove the „with“ parts?

jannikmi:
@Mourishitz please have a look again.

Mourishitz:
Sure, just noticed the problem here..

We now have two options, the test has passed even when GitHub actions said it was unexpected, so I don't think it is necessary anymore and we can remove this. Or, if really necessary we can try to pass it like

`--token=${{ secrets.GITHUB_TOKEN }}`

but, as I don't have enough knowledge about the impacts of this chance, I would like to discuss before commiting this changes. I do think removing is a better option because I also can't test it locally, so I don't know what to expect when running --token inside the 'extra_args' block

jannikmi:
@Mourishitz Thanks for your contribution!

PR #205, opened 2023-06-15
Correct tzfpy docs

ringsaturn:
It's true that tzfpy is not so accurate around bounders(about 3km ranges). But it is because of using simplified polygons.

tzfpy uses rectangles as index to speed up but auto fall back to polygon data if cache is missed. It's written in Rust here: https://github.com/ringsaturn/tzf-rs/blob/v0.4.1/src/lib.rs#L307-L318

### DIFF
--- a/docs/3_about.rst
+++ b/docs/3_about.rst
@@ -74,11 +74,11 @@ Comparison to tzfpy
 
 - ``tzfpy`` is a Python binding of the Rust package ``tzf-rs``
 - ``tzfpy`` has no startup time
-- ``tzfpy`` uses a hierarchical tree of rectangles as timezone polygons (data):
+- ``tzfpy`` uses simplified timezone polygons (data):
     - this reduces the memory requirements
     - this reduces the accuracy
     - this increases the lookup speed
-
+- ``tzfpy`` uses hierarchical tree of rectangles to speed up the lookup but auto fall back to polygon data if cache miss
 
 Comparison to pytzwhere
 -----------------------


ringsaturn:
> Cool. What is the data source and size of the simplified timezone polygons? The data source currently used by `timezonefinder` is quite huge (even when optimised, https://github.com/evansiroky/timezone-boundary-builder/releases/tag/2022d).

Here are things has been done:

- Polygon data has been convert to Protocol Buffers, which has a good size compress result for dumping to disk(nothing change on memory usage)
	- Proto define: https://github.com/ringsaturn/tzf/blob/main/pb/tzinfo.proto
	- Process pipeline: https://github.com/ringsaturn/tzf-rel/blob/main/.github/workflows/ci.yml#L58-L59
- Use Douglas Peucker to simplify polygon points: file size from 85MB -> 11MB; memory usage from 900MB to 60MB(for tzf written in Go) or 40MB(for tzf-rs and it's Python&Ruby binding)
	- https://github.com/ringsaturn/tzf/blob/main/reduce/reduce.go#L18
- When packaging to Python packages, another zip compress will be done, which reduce whl size(about 7MB)
	- https://peps.python.org/pep-0427/
	- https://pypi.org/project/tzfpy/#files

I have wrote a [post](https://blog.ringsaturn.me/en/posts/2023-01-31-history-of-tzf/) about history of tzf and it's Rust/Python binding, which has more details about those chanegs.

PR #282, opened 2025-07-06
Move to uv and setuptools-based extension build

### LINKED ISSUE #279: Update dependency management
Following work in #267, it's reasonable to support stable Poetry 2.x

theirix:
Since building extension modules via `build.py` is [not officially supported by Poetry](https://python-poetry.org/docs/main/building-extension-modules#maturin-pyproject) even in Poetry 2.0, it's worth evaluating other build systems. The [uv](https://docs.astral.sh/uv) is a promising alternative, although the uv build backend doesn't support build scripts as well. But it is possible to use a setuptools backend while using a modern `pyproject.toml`-based tooling:
```toml
[build-system]
build-backend = "setuptools.build_meta"
```

Changes:
- Convert to a standard `pyproject.toml`
- Introducing a standard setuptools-based CFFI build, which requires `setup.py` (could be migrated in future years to a pure `pyproject.toml`-based build when stabilised).
- Move files to [suit flat layout](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#flat-layout) with setuptools.
- Replace `poetry` calls with equivalent `uv` calls, except in the Makefile (could be done later).

Closes #279


jannikmi (approved):

jannikmi:
Great! Thanks for your work!

PR #285, opened 2025-07-08
Use uv in Makefile invocations

theirix:
A follow-up for #282. 
All makefile snippets are also migrated to uv or generic pip.

I built and tested wheels for only of my applications, they work well. I'd be interested in testing GHA-built wheels, too. Could you please graft a pre-release?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,40 +5,38 @@ SHELL=/bin/bash
 install:
 	pip install --upgrade pip
 	@echo "installing all specified dependencies..."
-	@#poetry install --no-dev
 	# NOTE: root package needs to be installed for CLI tests to work!
-	@poetry install --all-extras --sync
+	@uv sync --all-groups
 
 update:
 	@echo "updating and pinning the dependencies specified in 'pyproject.toml':"
-	@poetry update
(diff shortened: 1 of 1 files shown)


jannikmi:
Thanks, @theirix 
I added the version 6.5.9rc1 and the GHA seems to run fine.
With the current GHA definition, one would need to merge the PR to actually create a release.
Should we split up the `publish` step into `publish GitHub` (run only on always) and `publish pyPI` (run only on main)?

theirix:
Ah, I forgot about Python 3.8, reverting to the original license change.

Also, with the introduction of the `setuptools.package` field, we need to specify the binary package too - fixed.

theirix:
@jannikmi pushed the changes.

After this PR is merged, we will see a new GH release and TestPyPI package.

After the master is tagged, we will see the uploaded sdist and wheel on PyPI.

theirix:
Works good. Verified TestPyPI, sdists and wheels are in a good state

```sh
docker run --rm -it -v `pwd`:/workspace/ python:3.11 bash -c "pip3 install /workspace/timezone*whl ; python3 -c 'from timezonefinder import TimezoneFinder; assert(TimezoneFinder.using_clang_pip())'"
```

Issue #312, opened 2025-07-21
use pydantic for validating and parsing the input data geoJSON


jannikmi:
use `pydantic` in `scripts/file_converter.py` to validate the expected format of the geoJSON input dataset.

Note: `pydantic` should only be added as a development dependency, since the file parsing script does not belong to the main package

nikkhilaaryan:
Hi, I am interested to contribute to this issue, may I work on it?

jannikmi:
Sure, go ahead!

PR #327, opened 2025-07-29
GeoJSON validation using Pydantic in file_converter.py

aryannikhil:
@jannikmi 
Hey!
This PR addresses #312 by adding GeoJSON input validation to file_converter.py using Pydantic. I created a GeoJSON schema in scripts/geojson_schema.py and integrated it into the parsing step. Also added a sample file under tests/test_input.geojson.

Open to any suggestions or changes ,happy to improve things as needed.
Thanks!

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ Changelog
 
 Internal:
 * using abi3 (aka Python limited API) wheels to avoid a combinatory explosion with Python version. It allows the use of a single Python 3.9 base and building future-proof wheels. Thanks to `theirix <https://github.com/theirix>`__
+* using pydantic to validating and parsing the GeoJSON dataset. thanks to `ARYAN RAJ <https://github.com/nikkhilaaryan>`__ for the PR.
 
 
 8.0.0 (2025-08-11)


uv.lock
jannikmi, line 1: pydantic is required as dev. dependency

scripts/geojson_schema.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Feature(BaseModel):
+    type: str
+    properties: FeatureProperties
+    geometry: Geometry
+
+
+class GeoJSON(BaseModel):
+    type: str
```
jannikmi, line 22: the `type` properties are not used and hence do not need to be validated

scripts/geojson_schema.py
```diff
@@ -0,0 +1,23 @@
...
+    coordinates: Union[list, List[List[float]], List[List[List[float]]]]
+
+
+class FeatureProperties(BaseModel):
+    tzid: str
+
+
+class Feature(BaseModel):
+    type: str
```
jannikmi, line 16: type should be validated to be exactly "Polygon":

`type: Literal["Polygon"]`

jannikmi (changes requested):

jannikmi:
@nikkhilaaryan thanks for this PR. Please have a looke at the comments I left in the review

Issue #328, opened 2025-08-01
drop global variable use in file_converter.py script


jannikmi:
Scope: `file_converter.py` script

Building on top of the changes of PR #327 the validated pydantic GeoJSON representation should be converted into a data structure (also pydantic base model) with all the properties required for storing the binary data files. This makes all global variables in the script obsolete

- [ ] pydantic class with all required properties (before: global variables)
- [ ] `.from_geojson()` `@classmethod` to construct instace from initial pydantic GeoJSON instance (#327)
- [ ] specific validators for the properties (coordinate numpy array shapes etc.)

jannikmi:
Yes, sure! Please Open a PR so all contributors can Review the changes.

I can highly recommend to use a coding agent (e.g. cursor, GitHub copilot...) for implementing the changes. They can take care of such small refactorings easily.

pratyushkumar211:
Thanks @jannikmi and also for the suggestion! I used Grok4 as the coding agent. I have submitted a PR https://github.com/jannikmi/timezonefinder/pull/341. Please take a look and I look forward to hearing from you on the changes or any feedback.

PR #341, opened 2025-09-14
ISSUE-328 Refactoring in file_converter.py to drop global variables and store data in a class

pratyushkumar211:
Based on the suggestions in https://github.com/jannikmi/timezonefinder/issues/328, I have modified file_converter.py with the following changes: 

- Created TimezoneData class to store all the global variables 
- parse_polygons_from_json now calls .from_geojson from TimezoneData
- Added validators for the properties
- All dependent functions that were affected by this change in file_converter.py are modified.


jannikmi (dismissed):

jannikmi:
@pratyushkumar211 thanks for your contribution! I ended up refactoring the file parsing quite a bit more, since the shortcut compilation was super slow. the optimised code is now super fast and of much higher quality. also due to you!

PR #346, opened 2025-09-17
294 precompute unique zone ids

### LINKED ISSUE #294: performance optimisation: precompute unique zone ids
~ 75% of the h3 hexagon cells used as spatial index, have 1 unique zone which could be immediately returned without any further lookup (cf. [data_statistics](https://timezonefinder.readthedocs.io/en/latest/data_report.html#shortcut-mapping-statistics))

- [ ] precompute and store a mapping from hexid to unique zone id in `flatbuffer` binary format similar to the shortcut mapping being used already (cf. [shortcut_utils.py](https://github.com/jannikmi/timezonefinder/blob/master/timezonefinder/flatbuf/shortcut_utils.py))
- [ ] immediately return the corresponding timezone name in case of a match
- [ ] consider deleting the polygons which are only used in shortcuts together with polygons of the same zone. these will never be checked with an actual PIP test. Attention: what is the expected behavior of `get_geometry()` in these cases?

jannikmi:
Closes #294


jannikmi:
The performance actually deteriorated with these changes (see performance benchmark results). leaving this open as a draft for future reference. perhaps these changes come in handy later.

jannikmi:
improved the performance further to actually reach superior speed. the optimisation was valid afterall

PR #348, opened 2025-09-18
256 precompute last zone change index of shortcut

### LINKED ISSUE #256: performance optimisation: precompute last zone change index of shortcut
Each Timezone Polygon corresponds to a zone id. 
Each shortcut (h3 hexagon) points to a list of polygon ids sorted after their zone ids. 
Currently during each lookup it is being computed where the list of sorted zone ids change for the last time (`last_zone_change_idx`). 
to speed up the lookups, this value can be precomputed and stored as part of the binary file compilation in `scripts/file_converter.py`

jannikmi:
benchmarks show: in practice the overall throughput is dominated by the point-in-polygon work.


jannikmi:
Closing this: half of it shipped by another route, and the other half is now measured and is not
worth what it costs.

**What shipped instead.** The unique-cell fast path this branch built a second FlatBuffers
collection for exists in the packaged data today — `hybrid_shortcuts_uint16.bin` stores a unique
cell's zone id *inline* as a union member, so the common case is answered without touching a
polygon list at all. A separate `unique_shortcuts` binary would now duplicate that.

**What is left, priced.** Precomputing `last_zone_change_idx` removes `get_last_change_idx`:
**149 ns on the numba path and 283 ns on clang, out of a ~13.3 µs ambiguous query, and nothing at
all on a unique-shortcut one.** Uniformly random points are ~25 % ambiguous, so that is **~1 % of a
realistic mixed workload** — under the 3–9 % run-to-run noise of the machine that would have to
demonstrate it, i.e. not something the benchmark suite could confirm even if it were real. Against
that: a shortcut-layout version bump, therefore a `DATA_FORMAT_VERSION` bump, therefore an ordered
two-distribution release (data first, then the code that requires it).

The 2025 verdict on #256 — *"throughput is dominated by the point-in-polygon work"* — was right.
It now has a number under it, from `prototypes/query_stage_profile.py` (#497), which attributes a
query to its stages per backend and per coordinate-access mode.

**The part that is still open.** The same block of code spends ~0.8 µs per ambiguous query in two
numpy calls over a handful of candidates — `zone_ids_of` at ~600 ns plus this one — which is more
than coordinate validation, the H3 cell lookup and the shortcut lookup put together. Narrowing the
candidates in one pass without the numpy round-trip needs **no format change at all**, and is
tracked as `PERF-2` in `potential-improvements.md`. That is where the remaining win is.

The branch is also a year behind the tree it targets: it writes `timezonefinder/data/`, which is now
the separate `timezonefinder-data` distribution, and predates the current `flatbuf/io` layout.

Recorded under *Recorded decisions* in `potential-improvements.md` (PR #540) so it is not proposed a
third time — including what shipped in its place, which is the part that would otherwise be
rediscovered.

PR #369, opened 2025-11-15
Fallible CFFI extensions builds

### LINKED ISSUE #359: Can’t pip install time zone finder under Python 13.3.7
Sorry last post had wrong number. 13.3.3 was ok but failed on pip install timezomefinser under 13.3.7

theirix:
Problem: We recently switched to PyPI binary wheels to ensure a C extension build. Initially, it was possible to optionally build a C extension and ignore any errors, which was fragile and unreproducible. However, we cannot require C build for everyone, because even in the presence of the `cffi` package, a compiler may be missing, even on a Linux system and very often on a Windows system.

Older approach in this PR:
~~Let's make C extension build mandatory for CI builds (so PyPI wheels contain the binary) and allow users to opt-in C extension build via `BUILD_CFFI` environment variable (say, when running `uv build` or `pip wheel`). Unfortunately, we cannot use package extras to control it, so an environment variable is the only viable option.~~
It didn't work due to the significant complexity of the tox build and test sandboxing.

Solution:
- enable a fallible C extension build, so users won't worry about enabling it
- add end-to-end test to verify that wheels are built with binary extension
- fix tox tests to run examples only for pytz extra in rare cases of limited tox invocation
- refactor [old-style](https://docs.python.org/3/library/subprocess.html#legacy-shell-invocation-functions) `subprocess.getoutput` call

Closes #359

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,2730 @@
+version = 1
+revision = 2
+requires-python = ">=3.9, <4"
+resolution-markers = [
+    "python_full_version >= '3.12'",
+    "python_full_version == '3.11.*'",
+    "python_full_version == '3.10.*'",
+    "python_full_version < '3.10'",
+]
+
+[[package]]
+name = "alabaster"
+version = "0.7.16"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/c9/3e/13dd8e5ed9094e734ac430b5d0eb4f2bb001708a8b7856cbf8e084e001ba/alabaster-0.7.16.tar.gz", hash = "sha256:75a8b99c28a5dad50dd7f8ccdd447a121ddb3892da9e53d1ca5cca3106d58d65", size = 23776, upload-time = "2024-01-10T00:56:10.189Z" }
+wheels = [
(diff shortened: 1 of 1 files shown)


uv.lock
jannikmi: @theirix: why do you want to completely remove the lock file from version control?

I was aiming at this behavior:
> Commit the lock file to source control for library development, but do not include it in the published build/package.
> This gives you the best of both worlds: deterministic development and flexible consumers.
> 
> 1. Reproducible development and CI
> 
> The lock file ensures:
> 
> Contributors all use the same dependency versions.
> 
> CI runs tests against a stable, known dependency tree.
> 
> Bugs caused by upstream updates don’t appear randomly.
> 
> This improves reliability when developing the library.
> 
> 2. But you should not include the lock file in published library artifacts
> By excluding the lock file from the build, downstream users resolve dependencies normally using version ranges.

jannikmi:
@theirix would you mind keeping the `uv.lock`?

theirix:
> @theirix would you mind keeping the `uv.lock`?

I have a feeling that libraries shouldn't include `uv.lock`, but end applications should. Do we have a special need for it?

jannikmi:
@theirix 
I was aiming at this behavior:

Commit the lock file to source control for library development, but do not include it in the published build/package.
This gives you the best of both worlds: deterministic development and flexible consumers.

Reproducible development and CI
The lock file ensures:

Contributors all use the same dependency versions.

CI runs tests against a stable, known dependency tree.

Bugs caused by upstream updates don’t appear randomly.

This improves reliability when developing the library.

But you should not include the lock file in published library artifacts
By excluding the lock file from the build, downstream users resolve dependencies normally using version ranges.

Issue #317, opened 2025-07-22
reduce release memory footprint


jannikmi:
Hitting the project size limit of 10 GB.
Had to file an issue to request for more project space: https://github.com/pypi/support/issues/7038
Had to delete all pyPI releases up to version ``3.4.2`` (last version supporting python 2.7) to free up project space

> How large is each release?
> 
> 600 MB in Total
> 
> The current footprint of one distribution is around 51 MB.
> 11 distributions per release: 1 source, 10 wheels

- how can we reduce the release size?
- do we really need (10) wheels?
- is it fine to build wheels only for 1 python version?
- is it fine to build wheels only for 1 linux version?
- are there any other ideas?

theirix:
1. New wheel size

I downloaded the latest 7.0.0 release wheel for Linux (timezonefinder-7.0.0-cp311-cp311-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl) The size of a native binary is 32K, which is approximately the size of the `timezonefinder.py` file (22K).
99.7% of the package size is geo data - 62 Mb. Compressed wheel size is 51Mb, uncompressed is 62Mb - zip saves 14%.

2. Old wheel size

For older (6.5.9) one-geo-data-file wheels, the size of the geo data is 59 Mb of compressed wheel size 49 Mb. So the old package is slightly smaller.

3.  Geodata size

Even flatbuffer data can be compressed. A simple tar with lzma (`tar cf - data | xz --best | wc -c`) gives 22 Mb instead of 62 Mb. Of course, wheel's zip compression won't help now, and the wheel stays the same.

Hence, the savings with this approach are 60%.

The low-hanging fruit - use npz instead of npy (but it's only 40Kb). The largest file is the flatbuffer file. Simply unpack with the built-in `lzma` compression while reading,

4. Do we really need (10) wheels?

That's a good question. I would provide several of the latest versions for all Linux/Python types. Older versions could be stripped from PyPi with a GHA cron job. Still, it's worth attaching wheels to GitHub releases in case people want to get them. It's entirely reasonable to download them from GitHub releases, even programmatically.

5. musl vs glibc Linux flabours

Generally, musl is quite helpful for lightweight usage. For example, I use it in containers and AWS images

jannikmi:
> Geodata size
> 
[...]

Issue #293, opened 2025-07-14
timezone binary data compression


jannikmi:
This package aims to be lightweight, but with the current timezone data the disk footprint is > 60 MB (wheel size around 51 MB).
Look into techniques for compressing the timezone data (binary files) without loosing precision or speed.

Idea: research the effect of using 16 bit integegers instead of 32bit. what is the worst case precision at the equator?
Note: there are other packages like [tzfpy](https://github.com/ringsaturn/tzfpy) which are faster, but offer less accuracy. reducing the accuracy in `timezonefinder` probably diminishes the USP of this library (cf. https://github.com/ringsaturn/tzfpy/issues/94)

jannikmi:
Design notes carried over from #392 (closed unfinished), so the next attempt does not start from scratch.

### The size win is real

#392 measured roughly 60% off the wheel by zstd-compressing the FlatBuffers binaries, consistent with @theirix's lzma estimate in #317. A wheel is already a deflate zip, but deflate only buys ~14% on this data, so this is a genuine additional gain rather than double counting.

### Blocking finding: "decompress once at install time" is not possible with wheels

My review of #392 asked for decompression to happen once at install time, with the runtime read path left untouched and reading plain uncompressed files. That turns out not to be implementable: **a wheel has no install-time hook**. `pip` unpacks the zip into `site-packages` and runs no project code — that is a deliberate property of the wheel format, not a gap that will be closed. Only an sdist install runs a build backend, and requiring an sdist install would force a compiler on every user for the C extension.

So the decision an implementation has to make is *when* decompression happens, and every remaining answer has a real cost:

- **Lazy unpack on first use.** Needs `timezonefinder/data/` (or a fallback cache dir) to be writable — it is not in read-only container images or system-managed installs — plus an atomic, concurrency-safe unpack, since simultaneous cold starts of a service is exactly our use case. First call pays the decompression latency, and the compressed copies have to be deleted afterwards or the on-disk footprint grows rather than shrinks.
- **Explicit unpack step** (a console script or CLI subcommand). No magic and no writability surprises, but it breaks `pip install` and go, so it needs a precise error at lookup time when the data has not been unpacked yet. `TimezoneFinder(bin_file_location=...)` already exists as the hook for pointing at an unpacked directory elsewhere.
- **Decompress into memory at load.** Rules itself out on our own constraints: it breaks the memory-mapped `in_memory=False` path that keeps this usable in memory-constrained containers, and it charges every process a startup cost.

[...]

PR #392, opened 2026-01-26
successfully implemented FlatBuffer data compression for the timezone…

crazy4devops:
…finder project to address issue #317 (reduce release memory footprint). Here's what was accomplished:

Wheel size reduced by 60%: 51 MB → 18-20 MB
Release size reduced by 64%: 560 MB → 200 MB
Storage capacity increased 2.7×: Can now store 50+ releases instead of 18 Solves the storage crisis without deleting existing releases

### DIFF
--- a/CHANGES_SUMMARY.md
+++ b/CHANGES_SUMMARY.md
@@ -1,0 +1,245 @@
+# Quick Reference: All Changes Made for FlatBuffer Compression
+
+## 1. New File: `timezonefinder/flatbuf/io/compression.py`
+
+A complete compression/decompression module for zstandard (zstd) support.
+
+**Key Functions:**
+- `compress_file(input_path, output_path)` - Compress file to disk
+- `decompress_file(input_path, output_path)` - Decompress file to disk  
+- `decompress_bytes(data)` - Decompress bytes in memory
+- `decompress_mmap(compressed_mmap)` - Handle decompression from mmap
+
+**Configuration:**
+- Compression level: 19 (maximum, optimized for geodata)
+- Uses zstandard (zstd) library
+
+---
+
+## 2. Updated: `timezonefinder/flatbuf/io/polygons.py`
+
+### Added Imports:
+```python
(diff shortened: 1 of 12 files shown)


jannikmi:
Thank you for this PR.

The way I understand the implementation, currently the compressed files will be uncompressed on demand, so every time the binary files will be opened (`in_memory=False`). Is that correct?
I wonder how much overhead this adds to every call. Has this been evaluated?


Why has `zstandard` been chosen for compression?

I have a couple of open TODOs:

- [ ] uncompression on install. client code (read) should only still support uncompressed files
- [ ] remove compressed files after uncompressing. avoid avoid larger storage footprint when storing both file versions
- [ ] adding relevant design decisions to this discussion (uncompressing on demand,...)
- [ ] are there alternatives to the `zstandard` dependency used for compression? what are the tradeoffs?
- [ ] removing AI generated .md files
- [ ] run pre-commit hooks
- [ ] fixing tests
- [ ] fix CI/CD runs
- [ ] adding compressed binary files to version control (replacing uncompressed files)
- [ ] adding relevant documentation

jannikmi (changes requested):

jannikmi:
@crazy4devops do you have time to work on this further?
Do you need clarification with any of the remaining work items?

jannikmi:
Thanks again for taking this on.

I'm closing this as it's gone stale and conflicts with master by now. This isn't a rejection of the approach — compressing the FlatBuffers data is something I still want, and your PR showed the win is real (~60% off the wheel). The blocker is architectural: this branch decompresses on demand at read time, whereas the direction I'd want is decompression once at install time, with the runtime read path unchanged and uncompressed. That's a rewrite rather than a revision of this diff, so it makes more sense to restart from the design than to keep this open.

I've summarised what we learned here in https://github.com/jannikmi/timezonefinder/issues/293. If you'd like to pick it up again from that design, you're very welcome — just say so on the issue.

The full diff stays available at refs/pull/392/head (930f18ba) for anyone who wants to reference it.

PR #410, opened 2026-04-12
Using python 3.10 type hint features

### LINKED ISSUE #389: Use python 3.10 syntax throughout
Python 3.9 is not officially supported any more. 
Use 3.10+ syntax (like simplified typing: str | None)

aureliobarbosa:
This PR updates the repository using typing notation introduced in python 3.10.  Legacy typing module constructs (Dict, List, Set, Tuple) where exchanged for the native syntax (dict, list, set, tuple) and the union operator | is used to  avoid using Optional and Union. In addition to that, TypeAlias is used explicitly on type variables.

Initial investigation was done by reading [python's whatsnew](https://docs.python.org/3/whatsnew/3.10.html) page and an original plan was done with ChatGPT 5.3. The execution of the plan was done using Claude Sonnet 4.5 and this refactoring was checked in detail, changed by change. Prompts can be made available under request. I am available for any question.

Fixes #389

### DIFF
--- a/prototypes/single_resolution_bench.py
+++ b/prototypes/single_resolution_bench.py
@@ -36,7 +36,7 @@ from collections import defaultdict
 from dataclasses import dataclass
 from functools import lru_cache
 from pathlib import Path
-from typing import Any, Sequence, Union
+from typing import Any, Sequence
 
 import h3.api.numpy_int as h3
 import numpy as np
--- a/scripts/benchmark_utils.py
+++ b/scripts/benchmark_utils.py
@@ -9,7 +9,7 @@ to generate RST reports and handle CLI interfaces.
 import argparse
 import platform
 from pathlib import Path
-from typing import List, Dict, Any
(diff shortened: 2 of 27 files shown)


aureliobarbosa:
I will fix these issues. It seems that tests are annotated with types Dict, etc. And that the AI missed them because it used a regex to search for `Dict[`. I am converting to draft and will review it. What I think is funny is that all hooks and tests passed on my machine! So, something went wrong and it points to the need of even more careful review on the work done with the AI help...

jannikmi:
@aureliobarbosa no worries.
Make sure to activate the pre-commit hooks by running `make hook`

jannikmi (dismissed):

aureliobarbosa:
@jannikmi Done here. This final step was AI-less.

jannikmi (approved):

jannikmi:
@aureliobarbosa thanks for your work!

PR #413, opened 2026-04-16
Modernize codebase with Python 3.11+ features

jannikmi:
migrate typing imports, add Self type annotation, and refactor conditional dispatches to match/case statements


jannikmi:
@aureliobarbosa following up on your work

Issue #364, opened 2025-10-21
Enable Free-threaded Python (FTP)


jannikmi:
https://py-free-threading.github.io/

aureliobarbosa:
I did some investigation on this.

Looking into the dependencies of this project, keeping in mind free-threaded python version 3.14, the situation is the following:

```
dependencies = [
    "numpy>=2",            # Support is improving on each version. Numpy 2.4 supports python 3.11-14
    "h3>=4",               # Status unknown, nothing is mentioned on h3-py repository
    "cffi<3,>=1.15.1",     # Supports Python 3.14 free-threaded on version 2.0.
    "flatbuffers>=25.2.10",# Doesn't mention.
]
```
Considering additional dependencies, installing numba comes with problems. Only the latest numba (Version 0.65.0, 31 March 2026) release accepts free-threaded python 3.14. So, only recently people got it to work on this package. Also, when I installed numba version 0.65 on my environment using uv, the tool automatically selected numpy version 2.4.4 (which is the latest one).

While running the complete test suit pytest complains about h3 not declaring that it can run on free-threaded python:

```bash
❯ uv run make testall
platform linux -- Python 3.14.4, pytest-9.0.3, pluggy-1.6.0
**<frozen importlib._bootstrap>:491
  <frozen importlib._bootstrap>:491: RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'h3._cy.cells', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.**
=========================== 2341 passed, 2 skipped, 1 warning in 110.63s (0:01:50) ============================
```
But it also indicates that it is possible to force execution (at your own risk!) using `PYTHON_GIL=0`... 

By doing this all regular tests passes. It doen't mean much, since it should be necessary to implement specific parallel tests. 

```bash
timezonefinder on  free_thread_python [!] is 📦 v8.2.2 via 🐍 v3.14.4 (timezonefinder) 
❯ PYTHON_GIL=0 uv run make testall
============================================= test session starts =============================================
platform linux -- Python 3.14.4, pytest-9.0.3, pluggy-1.6.0
configfile: pyproject.toml
collected 2343 items                                                                                          

tests/cli_test.py ......                                                                                [  0%]
tests/global_functions_test.py ........................................................................ [  3%]
....................................................................................................... [  7%]
....................................................................................................... [ 12%]
[...]

PR #416, opened 2026-04-21
fixes I/O overlap on multithread tests using temp_file_path

aureliobarbosa:
While running this repo with pytest-run-parallel pytest plugin it was raising three multithread errors on the tests on file  
`tests/test_optimized_hybrid_shortcuts.py` (test_write_read_roundtrip, test_empty_data, test_single_element_arrays_should_not_occur). 

The problem was that the fixture `temp_file_path` was non thread safe, pytest per se is not adapted to multithread and the fixtures are collected before running the tests. The workaround is to generate a file factory, and tag the file with the thread. This way each thread get its own unique file to do the test. 

I decided not to change tox configurations because pytest-run-parallel has a deprecation warning on a feature that would allow for only running  tagged functions as parallel. It looks like this tool prefers users to run the whole test in parallel by default. So a decision must be made about this, but pytest-run-parallel will possibly change or improve in the near future.


aureliobarbosa:
related to #364.

PR #431, opened 2026-07-15
test: add hypothesis property tests for coordinate validation (#143)

### LINKED ISSUE #143: add property tests (hypothesis)
https://hypothesis.readthedocs.io/en/latest/quickstart.html
can be taken as baseline: https://github.com/jannikmi/multivar_horner/blob/master/tests/hypothesis_tests.py

```python
from hypothesis import given, strategies as st

@given(st.floats(-180, 180), st.floats(-90, 90))
def test_coordinate_validation_properties(lng, lat):
```

01luyicheng:
Closes #143

## Summary

Adds property-based tests with [`hypothesis`](https://hypothesis.readthedocs.io/) for `validate_coordinates`, `validate_lat`, and `validate_lng`, complementing the existing example-based `@pytest.mark.parametrize` tests in `tests/utils_test.py`.

The property tests explore the full finite input space:
- valid in-range coordinates pass through unchanged as floats
- out-of-range longitude/latitude raise `ValueError`
- `NaN` and infinity values raise `ValueError`
- `validate_lat` / `validate_lng` independently reject invalid input

This follows the snippet and baseline pattern suggested in the issue (https://github.com/jannikmi/multivar_horner/blob/master/tests/hypothesis_tests.py).

## Changes

- add `hypothesis>=6` to the `test` dependency group in `pyproject.toml`
- add `tests/test_property_validation.py` (7 property tests, all marked `@pytest.mark.unit`)
- add a changelog entry under the unreleased `Internal:` section

No source code under `timezonefinder/` is modified — this is a test-only change.

## Verification

- `uv run pytest tests/test_property_validation.py -v` → 7 passed
- `uv run pytest -m "not integration and not slow" -q` → 760 passed, no regressions
- `make hook` (all 22 pre-commit hooks: ruff, ruff-format, mypy, validate-pyproject, check-manifest, rstcheck, ...) → all passed

The new tests are pure-function (no dataset loading), fast (~0.9s total), and deterministic under the CI profile.

## Notes on the strategies

The `st.floats` strategies explicitly set `allow_nan=False, allow_infinity=False` for the "valid" and "out-of-range" cases so that hypothesis does not emit `NaN`/`Inf` where they are not expected. `NaN` and `Inf` are exercised via dedicated strategies with the appropriate flags. Boundary values (`±180` longitude, `±90` latitude) are inclusive and treated as valid, matching the `<=` comparisons in `is_valid_lat` / `is_valid_lng`.

### DIFF
--- a/tests/test_property_validation.py
+++ b/tests/test_property_validation.py
@@ -5,8 +5,6 @@ Uses ``hypothesis`` to exercise the full finite input space of
 complementing the example-based tests in ``tests/utils_test.py``.
 """
 
-import math
-
 import pytest
 from hypothesis import given, strategies as st
 
@@ -31,10 +29,10 @@ _OUT_OF_RANGE_LAT = st.floats(allow_nan=False, allow_infinity=False).filter(
     lambda x: not (-90.0 <= x <= 90.0)
 )
 
-# NaN values (allow_nan=True; the filter keeps only the NaN draws).
-_NAN = st.floats(allow_nan=True, allow_infinity=False).filter(math.isnan)
-# Infinity values (allow_infinity=True; the filter keeps only the Inf draws).
-_INF = st.floats(allow_nan=False, allow_infinity=True).filter(math.isinf)
+# NaN / Inf are only a few distinct values, so construct them directly instead
+# of generating many floats only to filter most of them away.
+_NAN = st.just(float("nan"))
+_INF = st.sampled_from([float("inf"), float("-inf")])
 
 # Pairs where at least one coordinate is NaN (resp. Inf); the other is valid.
 _NAN_PAIR = st.one_of(
--- a/tests/test_property_validation.py
+++ b/tests/test_property_validation.py
@@ -1,0 +1,109 @@
+"""Property-based tests for coordinate validation functions.
+
+Uses ``hypothesis`` to exercise the full finite input space of
+``validate_coordinates``, ``validate_lat``, and ``validate_lng``,
+complementing the example-based tests in ``tests/utils_test.py``.
+"""
+
+import math
+
+import pytest
+from hypothesis import given, strategies as st
+
+from timezonefinder import utils
+
+
+# Finite in-range coordinates. allow_nan=False and allow_infinity=False are
+# mandatory: otherwise hypothesis would emit NaN/Inf, which are invalid, and
+# the "valid" properties below would fail.
+_VALID_LNG = st.floats(
+    min_value=-180.0, max_value=180.0, allow_nan=False, allow_infinity=False
+)
+_VALID_LAT = st.floats(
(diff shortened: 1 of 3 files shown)


tests/test_property_validation.py
```diff
@@ -0,0 +1,109 @@
...
+# Finite floats strictly outside the valid range (bounds are inclusive).
+_OUT_OF_RANGE_LNG = st.floats(allow_nan=False, allow_infinity=False).filter(
+    lambda x: not (-180.0 <= x <= 180.0)
+)
+_OUT_OF_RANGE_LAT = st.floats(allow_nan=False, allow_infinity=False).filter(
+    lambda x: not (-90.0 <= x <= 90.0)
+)
+
+# NaN values (allow_nan=True; the filter keeps only the NaN draws).
+_NAN = st.floats(allow_nan=True, allow_infinity=False).filter(math.isnan)
```
jannikmi: Filtering works, but it's not ideal in Hypothesis because it generates many values only to discard most of them. Since there are only a few NaN/Inf values, you can construct them directly.

what about:

_NAN = st.just(np.nan)

tests/test_property_validation.py
```diff
@@ -0,0 +1,109 @@
...
+# Infinity values (allow_infinity=True; the filter keeps only the Inf draws).
+_INF = st.floats(allow_nan=False, allow_infinity=True).filter(math.isinf)
```
jannikmi: and:

_INF = st.sampled_from([np.inf, -np.inf])

jannikmi (changes requested):

jannikmi:
@01luyicheng thank you for your high quality contribution!

Are you willing to also add same kind of hypothesis tests for [all functions of the public API](https://timezonefinder.readthedocs.io/en/latest/4_api.html) in another test module?

01luyicheng:
Hi @jannikmi, thanks for the review!

I have applied your suggestion in a443348 — `_NAN` and `_INF` now construct the values directly with `st.just(float("nan"))` and `st.sampled_from([float("inf"), float("-inf")])` instead of filtering away most of the generated floats. I used `float(...)` rather than `np.nan`/`np.inf` to keep the module free of an extra import, but the behavior is identical. All 7 tests still pass.

And yes, I am happy to add hypothesis property tests for the public API functions (`timezone_at`, `timezone_at_land`, `unique_timezone_at`, `certain_timezone_at`, `get_geometry`) in a separate test module and PR. I will open that as a follow-up so this PR stays focused on the coordinate-validation properties.

01luyicheng:
Hi @jannikmi, just rebased this branch on the latest `master` to resolve a merge conflict in `CHANGELOG.rst` (both branches had added entries to the "Internal" section — kept both). The 7 property tests still pass. The review feedback from earlier is already applied in the second commit. Happy to make further adjustments if needed.

jannikmi (approved):

PR #433, opened 2026-07-16
test: add hypothesis property tests for public API (#143)

01luyicheng:
Follow-up to #431 as requested by @jannikmi: hypothesis property tests for the public API functions, in a separate test module.

## Scope

Adds `tests/test_property_api.py` with property-based tests for every function in the [public API](https://timezonefinder.readthedocs.io/en/latest/4_api.html):

- `timezone_at`
- `timezone_at_land`
- `unique_timezone_at`
- `certain_timezone_at`
- `get_geometry`

## Properties covered

**Return-value validity** (for each of the four lookup functions):
- Valid finite in-range coordinates return either `None` or a known timezone name from the canonical list.

**Error handling:**
- Invalid longitude (out-of-range, NaN, ±Inf) raises `ValueError` for every lookup function.
- Invalid latitude (out-of-range, NaN, ±Inf) raises `ValueError` for every lookup function.

**Cross-function relationships:**
- `certain_timezone_at` result, when not `None`, agrees with `timezone_at`. (The reverse direction does not hold near the poles, where the shortcut can still resolve a name that the exhaustive check misses — only the implication direction is asserted.)
- `timezone_at_land` result, when not `None`, agrees with `timezone_at` (land hits are a subset of all hits).
- `unique_timezone_at` result, when not `None`, agrees with `timezone_at`.

**Determinism:**
- Repeated calls with the same coordinates return identical results for all four lookup functions.
- `get_geometry` returns identical structures across repeated calls.

**`get_geometry`:**
- Every known timezone name has a non-empty polygon geometry (both coordinate-pair and list formats).
- Looking up by `tz_id` returns the same geometry as by name.

## Notes

- NaN/Inf values are constructed directly (`st.just`, `st.sampled_from`) rather than filtered, matching the approach agreed in #431.
- Invalid longitude and latitude use separate strategies because a value like `91.0` is a valid longitude but an invalid latitude.
- Example budgets are capped (`max_examples=50` for lookups, `25` for geometry) since polygon-in-point checks are not cheap.

All 19 new tests pass locally alongside the existing suite.


tests/test_property_api.py
```diff
@@ -0,0 +1,197 @@
...
+
+# A valid timezone name drawn from the known list.
+_TIMEZONE_NAME = st.sampled_from(sorted(_VALID_NAMES))
+
+# Keep the example budget modest: polygon lookups are not cheap.
+_LOOKUP_SETTINGS = settings(
+    max_examples=50,
+    deadline=None,
+    suppress_health_check=(HealthCheck.too_slow,),
+)
```
jannikmi: the real power of hypothesis comes from exhaustive test coverage. instead of limiting `max_examples`, please mark all tests in this module as slow:
```python
# Mark all tests in this module as slow
pytestmark = pytest.mark.slow
```

You can keep the `_LOOKUP_SETTINGS` for conveniently reducing `max_examples` for development, but in the production code please simply set `max_examples=None`.
Please apply `_LOOKUP_SETTINGS` to all test functions (also for `get_geometry`)

jannikmi (changes requested):
Thanks for the second, great PR!
I suggest some minor edits

jannikmi:
@01luyicheng somehow the CICD tests in this second PR are failing due to missing `hypothesis` dependency. Please have a look

01luyicheng:
@jannikmi Thanks for the detailed review! I've applied your suggestions:

1. **Added `hypothesis` to test dependencies** (`pyproject.toml` `[dependency-groups] test`) — this should resolve the CI failures
2. **Marked all property tests as `slow`** using `pytestmark = pytest.mark.slow` at the module level
3. **Set `max_examples=None`** in `_LOOKUP_SETTINGS` for exhaustive coverage in production, while keeping `_LOOKUP_SETTINGS` available for conveniently reducing `max_examples` during local development
4. **Applied `_LOOKUP_SETTINGS` to all `get_geometry` tests** (previously they used inline `@settings(max_examples=...)`)

The changes are pushed in commit 1fe107f. Looking forward to your next review.

jannikmi (dismissed):

jannikmi:
Sorry, for my wrong hint. `max_examples=None` is actually not allowed.
The tests are failing:
```bash
E   hypothesis.errors.InvalidArgument: Expected int but got max_examples=None (type=NoneType)
```

Please have a look how to use the default setting. I suggest simply commenting out the unused setting parameter lines.

01luyicheng:
Thanks for catching that! You're right — `max_examples=None` is not a valid value. I've commented out the `max_examples=None` line in `_LOOKUP_SETTINGS` so hypothesis falls back to its default, and updated the surrounding docstring/comments to match. All 19 property tests (plus the rest of the suite) pass locally now.

jannikmi (approved):

Issue #273, opened 2025-03-23
detect new available boundary data version


jannikmi:
add a new GHA workflow similar to this one: https://github.com/evansiroky/timezone-boundary-builder/blob/master/.github/workflows/tz_update_check.yml
It should check wether a new release is available here: https://github.com/evansiroky/timezone-boundary-builder/releases
open question: how can the data version currently used in this package be dertermined robustly?

nilsnolde:
just got here somewhat randomly from https://github.com/evansiroky/timezone-boundary-builder/issues/185. in our project we also do that with GHA. it's very specific to our project of course, but smth for inspiration: https://github.com/valhalla/valhalla/blob/master/.github/workflows/check_tz_releases.yml

hemkdev:
Hi @jannikmi! I'd like to take this one. 👋

Digging in, the current data version doesn't seem to live anywhere
machine-readable. `parse_data.sh` pulls from `releases/latest`, and the version
only shows up as prose in `CHANGELOG.rst`. So the check first needs a reliable
source of truth for the "current" version. I see two ways:

- **(a)** regex the top `CHANGELOG.rst` entry, for example matching `updated the data to 2026c`; no new code but fragile if the wording changes
- **(b)** add a `DATA_VERSION` constant to the package, for example `DATA_VERSION = "2026c"`; updated on release and more robust, but the release flow has to keep it in sync

I'd lean **(b)**. Which would you prefer before I start on the workflow?

jannikmi:
Thanks for your initiative!

I would suggest adding a file "DATA_VERSION" with the content being just "2026c" (currently)

Please refer to the previous comment from @nilsnolde for reference of an existing GHA workflow

hemkdev:
Perfect, will do!  Adding a `DATA_VERSION` file (`2026c`) plus the GHA check against `releases/latest`, using @nilsnolde's Valhalla workflow as a starting point.  PR coming soon. 🙌

volkkov:
Hit this same issue — the root cause was the missing null check before deserialization. Wrapping in a try/catch fixed it on our end.

jannikmi:
@volkkov sorry, I do not understand you comment. Was it meant for this issue?

Deniwn22:
Hit this same issue — the root cause was the missing null check before deserialization. Wrapping in a try/catch fixed it on our end.

jannikmi:
Thanks @hemkdev for your contribution!

[...]

Issue #167, opened 2022-10-30
script for automatically updating the boundary data version


jannikmi:
steps:
* create new branch
* download timezones-with-oceans.geojson.zip from https://github.com/evansiroky/timezone-boundary-builder/releases/latest
* unpack
* move .json to scripts folder
* execute file_converter.py
* update the existing binaries
* run ``tox`` and assert that all tests are passing
* bump version number in pyproject.toml (minor version bump)
* create new entry in changelog
* commit files
* raise Pull Request
* automatically merge if pipeline is running through

jannikmi:
note: the `parse_data.sh` script can be used as baseline

hemkdev:
Hi @jannikmi! Following up from #273 — I'd like to take this one too. 👋

Looking at the steps, I'd propose splitting it into three PRs, ordered by risk:

**PR 1 — make `parse_data.sh` CI-ready.** Replace the interactive `read` prompts with flags, auto-generate the changelog entry (the TODO already in the script), and clean up the version bump step. Fully testable locally, no workflow changes yet.

**PR 2 — automated update PR.** New workflow (triggered by the check from #273 instead of opening an issue): create a branch, run `make data`, commit, run the tests and open a PR. You'd still review and merge manually, but the manual effort drops to approving a ready-made PR.

**PR 3 — full automation.** Tag + auto-merge on green pipeline, creating the release. Kept last because it's the risky part (see question 2 below).

Two questions before I start:

1. **Version bump:** the issue says minor, but `parse_data.sh` currently does `uv version --bump patch` — and recent data-only releases (8.2.3, 8.2.5) were patch bumps. Which one should the automation use?
2. **Tag/auto-merge auth:** tags pushed with the default `GITHUB_TOKEN` don't trigger other workflows, so the tag wouldn't start the release pipeline in `build.yml`. Full automation needs a PAT or a GitHub App installation on the repo — only you can set that up. Fine to defer that to PR 3 and decide then?

If the split sounds good, I'll start with PR 1.

Deniwn22:
One thing worth trying: setting the timeout at the transport layer rather than the request level — avoids the race condition.

jannikmi:
Thank you @hemkdev 

Questions:
1. a patch version bump is fine
2. i will look into it. thanks for bringing it up.


about PR1:
- you could also translate the script to python if you think this is more convenient.
- feel free to pick a more suitable name for the script
[...]

PR #434, opened 2026-07-16
Open an automated update PR when new boundary data is available (issue #167, PR 2/3)

hemkdev:
Second PR of the plan discussed in #167, building on #432.

This extends the weekly `check_data_updates.yml` workflow (from #273): when a new timezone-boundary-builder release is detected, instead of only opening a notification issue, it now regenerates the data and opens a ready-to-review update PR.

**How it works:**
- The existing check job now exposes outputs (`update_needed`, `latest`, `current`) and skips when an update PR for that release is already open (dedup by branch name `data-update-<tag>`).
- A new `open_update_pr` job checks out with the GitHub App token (so the push and the PR trigger the regular CI/CD workflows), sets up all supported Python versions (the script runs the full tox suite), runs `update_data.sh --dataset=full --with-oceans --rm-tmp`, commits to a `data-update-<tag>` branch and opens a "Data `<tag>`" PR. The version bump and changelog entry come from the script itself (#432).
- If the automated update fails, a fallback job preserves the previous behavior and opens the notification issue, noting that manual regeneration is needed.

The PR-opening path can't be exercised end-to-end outside this repository (the `DATA_UPDATER_APP_ID`/`DATA_UPDATER_PRIVATE_KEY` secrets only exist here); the job logic was tested with stubs on a fork.

Part 3 (auto-merge + tag/release on green CI) follows once this lands.


.github/workflows/check_data_updates.yml
```diff
@@ -34,13 +38,122 @@ jobs:
...
+      # update_data.sh runs the full tox suite, which needs all supported interpreters
+      # NOTE: keep in sync with the versions in build.yml > tox.ini
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: |
+            3.11
+            3.12
+            3.13
+            3.14
```
jannikmi: Sorry for the inconvenience, but seeing this makes me realise that running tox within `update_data.sh` is redundant to the CI/CD pipeline also running the tests with the different python versions. Could you please remove the tox step from the script and simplify the python setup here?
Thanks

jannikmi (changes requested):
All looks good. I propose a small simplification

.github/workflows/check_data_updates.yml
hemkdev: Done @jannikmi, I removed the tox run from update_data.sh and simplified the workflow to a single Python version. Also adjusted the generated PR body accordingly: it now points to the update PR's own CI/CD pipeline as the validation step.

jannikmi (approved):

.github/workflows/check_data_updates.yml
jannikmi: Great work, Thanks!

PR #435, opened 2026-07-16
fix: update pull request permissions and improve existing PR check in…

jannikmi:
… workflow

### DIFF
--- a/.github/workflows/check_data_updates.yml
+++ b/.github/workflows/check_data_updates.yml
@@ -22,7 +22,13 @@ jobs:
       current: ${{ steps.check.outputs.current }}
     steps:
       - uses: actions/checkout@v7
-
+      - name: Test GitHub API
+        run: |
+          curl -v \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{ github.token }}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos/evansiroky/timezone-boundary-builder/releases/latest
       - name: Check for new release
         id: check
         env:
--- a/.github/workflows/check_data_updates.yml
+++ b/.github/workflows/check_data_updates.yml
@@ -22,29 +22,6 @@ jobs:
       current: ${{ steps.check.outputs.current }}
     steps:
       - uses: actions/checkout@v7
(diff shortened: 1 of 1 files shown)


jannikmi:
@hemkdev trying to fix the workflow here

hemkdev:
Thanks for looking into this @jannikmi! The missing `pull-requests: read` permission is a real bug on my side — the dedup `gh pr list` call would have failed with a 403 once it actually ran. Good catch.

Two observations though:

1. **The failure in the linked run was a different issue**: the job died on the *first* API call (`gh api repos/evansiroky/timezone-boundary-builder/releases/latest` — a public endpoint, so permissions don't affect it) with `invalid character '<' looking for beginning of value`, which is the GitHub API returning a transient 5xx HTML error page instead of JSON. The log shows the step died in ~0.4s, before even printing the data version. The upcoming PR 3 includes a retry (3 attempts) around these API calls to handle exactly this.

2. **Edge case with the branch-based check**: `git ls-remote` also matches a *leftover* branch. If a run pushes `data-update-<tag>` but fails before opening the PR, the next weekly run would see the branch and skip — blocking the update until someone deletes the branch manually. The PR-based check (now viable with the added permission) doesn't have this problem, since the `open_update_pr` job force-pushes over leftover branches. Alternatively, the branch check could stay if a leftover branch without an open PR gets deleted instead of skipped.

Happy to fold either variant into PR 3 if you prefer.

jannikmi:
@hemkdev please go ahead with your PR3 and the previous approach of checking for an open PR.

One note: even with extensive retries, I cannot get the fetch release call to work.

jannikmi:
@hemkdev I was able to fix the check using curl. have a look at the current version

hemkdev:
Nice, the explicit headers did the trick! Two small things before merging:

1. The two debugging steps ("Test GitHub API" and "Verify GitHub API headers") look like leftovers — they'll run on every weekly check.
2. Adding `--retry 3` to the curl call would also cover transient 5xx errors for free (`curl -f` + `set -e` still fails the job on a single hiccup).

Once this lands I'll rebase PR 3 on top and drop my now-redundant retry/permission commits — its scope goes back to just the auto-merge + release workflow.

PR #455, opened 2026-08-05
TEST ONLY: verify benchmark PR comment + alert (do not merge)

jannikmi:
Throwaway PR to exercise `benchmark-comment.yml`, which could not run before #454 landed on the default branch (workflow_run only fires for default-branch workflows).

Contains a deliberate `time.sleep(1e-5)` in `TimezoneFinder.timezone_at` so the comparison comment should show a large regression and trip the 110% alert threshold (non-blocking, `fail-on-alert: false`).

Will be closed and the branch deleted once the comment appears.


jannikmi:
Comment mechanism verified end-to-end; closing.

PR #493, opened 2026-08-10
Verify MANIFEST.in exclusions the other way round too

jannikmi:
## What

`tests/test_package_contents.py` asserts that nothing in the built sdist and wheel matches a pattern in `IGNORED_PATTERNS`. #486 closed one direction of that guard: a hand-written pattern naming no path in the checkout now fails.

The converse was still silent. `MANIFEST.in` and the pattern set are two hand-maintained statements of one intent, and they have drifted before — in both directions. An `exclude` / `recursive-exclude` / `prune` line added without a matching pattern leaves the file kept out **by the build and verified by nothing**: `test_no_unwanted_files_in_distribution` only ever asserts that *nothing* matched, so the day that line is deleted or stops matching, the file ships and the suite stays green.

Two new `unit` tests (no build needed, so a drifted exclusion surfaces in `make test`):

- **`test_every_manifest_exclusion_is_guarded`** — parses the `exclude` / `recursive-exclude` / `prune` / `global-exclude` directives out of `MANIFEST.in` and fails when one of them covers a path in the checkout that no pattern names. It asserts against the union `IGNORED_PATTERNS`, not `UNWANTED_DIST_PATTERNS` alone: several exclusions (`.claude/*`, `__pycache__/`) are already covered by a `.gitignore` pattern, and demanding a hand-written duplicate would only give the two copies a new way to disagree.
- **`test_pattern_only_exclusions_stay_current`** — keeps the exemption table below honest.

The four directive forms collapse into one shape (`prefix`, glob, `anchored`), and the glob is matched one path component at a time so `*` does not cross a separator — as it does not in `MANIFEST.in`, but does in `matches_pattern`.

Not attempted: deriving `UNWANTED_DIST_PATTERNS` from `MANIFEST.in`. Most of that set (`docs/`, `scripts/`, `benchmarks/`, `Makefile`, `tox.ini`, `uv.lock`, …) is excluded by setuptools *default* rather than by any `MANIFEST.in` line, so a derivation cannot replace it. This verifies agreement in one direction.

## The exemptions, and why

Four directives name untracked artefacts whose presence is a property of the machine rather than of the project, so scanning their paths would pass on one checkout and fail on another:

| directive | why it cannot be scanned | pattern carrying the same intent |
|---|---|---|
| `prune .git` | a directory in a clone, a **file** in a linked worktree | `.git/*` |
| `prune tests/.pytest_cache` | need not exist at all | `*.pytest_cache` |
| `global-exclude __pycache__/*` | holds numba's `.nbi`/`.nbc` only after a numba-enabled run | `__pycache__/` |
| `global-exclude *.pyc` | byte-code, never tracked | `*.py[cod]` |

These are exempted **against a named pattern**, not skipped: `test_pattern_only_exclusions_stay_current` fails if the directive leaves `MANIFEST.in` or the pattern leaves `IGNORED_PATTERNS`, so the exemption cannot outlive either side. This follows the existing `PATTERNS_WITHOUT_PROJECT_FILES` precedent in the same module.

I checked that neither `global-exclude` is load-bearing before exempting it: **dropping `global-exclude __pycache__/*` and rebuilding changes neither distribution** — setuptools already prunes `__pycache__` from an sdist by default. So these are defensive duplicates, not the only thing keeping byte-code out, and the numba cache files under `timezonefinder/__pycache__/` that a filesystem scan surfaces are not a real packaging hole.

## Mutation check

A guard I have not seen fail is not verified. Three mutations, each reverted:

1. Added `exclude SOMETHING_NEW.md` to `MANIFEST.in` with the file present → `test_every_manifest_exclusion_is_guarded` fails naming it:
   `MANIFEST.in excludes paths that no pattern in this module names (…): {'exclude SOMETHING_NEW.md': ['SOMETHING_NEW.md']}`
2. Removed `.cursor/` from `UNWANTED_DIST_PATTERNS` (the historical shape of the bug) → fails naming `.cursor/rules/repo-instructions.mdc`.
3. Removed `prune .git` from `MANIFEST.in`, and separately `__pycache__/` from `.gitignore` → `test_pattern_only_exclusions_stay_current` fails on the stale exemption and on the unbacked pattern respectively.

The test also asserts that the parser produced any directives at all, since an empty parse would make every check below vacuous — the same failure mode #486 fixed.

## Verification

- Both tests **pass on arrival**; no change to `MANIFEST.in` or to `UNWANTED_DIST_PATTERNS` was needed to make them pass.
- `pre-commit run --all-files` clean. (`make hook` itself cannot run in a linked worktree — `pre-commit install` refuses with `core.hooksPath` set — so the check half was run directly.)
- `make test` green; the new tests are selected there (`test_package_contents.py` contributes 3 unit tests).
- `make testint` green.
- `make testall` green — 2858 passed — run after `git fetch` and re-basing onto current `master`, in that order.
- Test-only: distributions are unchanged and no library code is touched. `git diff origin/master --stat` is `CHANGELOG.rst` + `tests/test_package_contents.py`.

## Notes

- This started stacked on #486, which merged while the work was in progress; it is now based directly on `master` and the changelog bullet for that PR is **amended** rather than duplicated, per `CLAUDE.md`'s one-bullet-per-change rule.
- Unrelated local snag worth knowing: `tests/test_integration.py::test_install_from_artifacts[wheel]` fails in a worktree whose `.venv` is older than the newest installed interpreter — `uv build` picks the newest (cp314 wheel) while `setup_venv` uses `sys.executable` (3.12). It reproduces unmodified on `master`. `UV_PYTHON=3.12 make testint` was used for the runs above.


jannikmi:
Rebased onto current `master` (`5a00273`), which brings in the documentation stack, and added the two `docs/architecture.rst` mentions:

- **Tests that protect guarantees, not behaviour** — a definition-list entry alongside the other four, on the angle that unites them: a check asserting *nothing matched* passes just as readily when its patterns match nothing at all.
- **How it ships** — the section covers the checks that stop a *broken* wheel being published (abi3 audit, the end-to-end install job) but not the one that checks what is *in* the artifact. The new paragraph states both halves and why the second is worth automating: a missing runtime file fails on first use and gets reported, an extra one ships quietly.

Two conflicts came out of the rebase, both resolved by keeping each side: `master`'s new #494 bullet next to the amended packaging-guard bullet in the changelog, and the docs stack's new prose after my definition-list entry in `architecture.rst`.

Re-verified on the new base: `pre-commit run --all-files` clean, `make docs` builds with no warnings, `make test` (1131) and `make testint` (375) green, `make testall` green at **2866 passed**. `make testint` no longer needs the `UV_PYTHON` pin mentioned in the description — #494 fixed that mismatch at the source.

PR #509, opened 2026-08-13
Store holes that duplicate a boundary polygon as a reference

### LINKED ISSUE #350: Replace holes with the corresponding boundary polygons
The dataset contains (boundary) polygons with holes. Hole are not really empty. It is a special property of the dataset that holes are actually filled by boundary polygons of another zone.

Storing and processing holes separately is actually redundant. This can be improved.

- [ ] First assert that for every hole exactly one boundary polygon of a different zone with equal bbox can be found
- [ ] ...

Also mention this in the upstream data set project?

jannikmi:
Implements `plans/350-holes-as-boundary-references.md`. Closes #350.

Opened fresh from `master` rather than rebasing `350-replace-holes-with-boundaries`: that branch's 100-commit-stale content is entirely `prototypes/` analysis the plan calls for discarding, and none of it is library code. **#351 is superseded and can be closed** — I have not touched it.

## What changed

Almost every hole is an enclave: the upstream builder cuts it into the surrounding zone using exactly the ring it also emits as the enclosed zone's own boundary polygon. On release `2026c`, **729 of 756 hole rings** trace the same closed path as some boundary polygon — the same geometry stored under two IDs.

The hole coordinate file now stores only the 27 rings without a twin. `holes/poly_ref.npy` records per hole id which boundary polygon to read instead (`v >= 0`), or where its own ring sits (`v < 0`, at `-(v + 1)`).

| | before | after |
|---|---|---|
| `holes/coordinates.fbs` | 2,111,472 B | 167,608 B |
| data directory (report) | 63.99 MiB | 62.14 MiB |

`in_memory=True` saves the same amount of RAM — those holes resolve into the boundary arrays instead of materialising a second copy.

Hole ids stay dense, so `hole_registry` and every caller above `coords_of()` is untouched, and the bbox vectors stay valid verbatim (a referenced ring is identical, so its bbox already equals the boundary's). `outside_bbox` — the hot path — keeps reading a flat array with no indirection.

## Corrections to the plan, and one judgement call

- **The plan's counts were measured against stale data.** Re-verified on `2026c`: 729/756 (96.4%) match, 0 near-misses, 0 ambiguous. Holes are 3.2% of coordinate data, not 5.5%. The exception set changed too — 27 not 19, `Asia/Hebron` covers the Jerusalem enclaves (not `Asia/Gaza`), and several ocean-zone cases are new.
- **"Byte-for-byte duplicate" was wrong for a third of the matches.** Only 463 of 729 are stored identically; 266 trace the same path from a different starting vertex or winding direction. You chose to match up to rotation/winding (1.84 MiB) over exact-only (1.08 MiB). Verified equivalent over **504,852 adversarial points** — every vertex and edge midpoint of the affected rings — on *both* PIP backends. The one visible consequence: `get_geometry()` may hand back those hole rings starting at a different vertex. Documented in the changelog.
- **The plan's build-time "matched boundary must belong to a different zone" assertion guards the wrong thing** and is a warning instead. The reference is exact whatever the zones are, so storing it changes no answer; hard-failing would block a data update over degenerate upstream geometry that predates this encoding. `tests/test_input.json` contains exactly that shape, so the assertion would have broken `make testparse`.
- **§6's open decision resolved by reuse:** `POLYGON_LAYOUT_VERSION` already exists on master (added after the plan was written) and does exactly what §6 asked for. Bumped to 2 so a released version rejects this data instead of resolving hole ids against the compacted file and returning plausible wrong answers. Layout 1 stays readable, so a `bin_file_location` directory compiled by an older release keeps working — the promise that marker was added to keep.

## Verification

- `tests/test_hole_references.py` (new): equivalence against a reconstructed all-inline dataset over every hole's interior plus 2k random points; both PIP backends on resolved rings; canonical-key properties; five rejection paths for inconsistent data directories.
- Packaged data regenerated from the pinned `2026c` release. `git status timezonefinder/data` listed only the two coordinate files and the new vector — every other binary reproduced byte-identically, which is what makes the lossless check meaningful.
- Full suite green (2884, including 1360 slow and 380 integration). Hole tests + `test_acceleration_paths.py` + `main_test.py` also run on a bare no-Numba env, where the C extension is the bound backend.
- Init and lookup timing neutral: ratio 1.004 (`in_memory=False`) and 0.995 (`True`).

Also fixes a pre-existing-shaped bug my change would have exposed: `PolygonArray.__del__` raised an unraisable `AttributeError` on a partially constructed instance, which `scripts/reporting.py` now creates when the layout guard rejects a file.

## Follow-ups (deliberately not here)

Dropping the 27 unmatched holes. They are fully covered by other zones — 0 of 1620 sampled interior points fell outside every other zone — so this is an *ordering* question, not a geometry one, and it needs an invariant on shortcut candidate order that does not exist yet. The rejected zone-precedence engine is recorded in one line of `prototypes/README.md` so it is not proposed again from scratch.


jannikmi:
@codex review

PR #518, opened 2026-08-17
Keep data-only releases from shipping pending work

### LINKED ISSUE #510: Automated data release reorders CHANGELOG.rst and tags work its notes don't describe
`update_data.sh` builds the changelog entry for a data update by splicing it in directly under the
three-line header (`update_data.sh:140-147`):

```sh
{
    # keep the changelog header (first 3 lines), insert the new entry below it
    head -n 3 "$CHANGELOG_PATH"
    printf '\n\n%s\n%s\n\n' "$ENTRY_TITLE" "$ENTRY_UNDERLINE"
    printf '* updated the data to `%s <%s/releases/tag/%s>`__\n' "$DATA_TAG" "$DATA_REPO_URL" "$DATA_TAG"
    tail -n +4 "$CHANGELOG_PATH"
} >"$CHANGELOG_PATH.new"
```

That is correct when `X.X.X (unreleased)` is empty, which is the case it was written for. It is not
the case that actually occurs: the weekly `check_data_updates.yml` fires whenever upstream publishes,
regardless of what has accumulated on `master` since the last release.

## Two things go wrong when the unreleased section is non-empty

**1. The file ends up out of order.** The new dated section lands *above* `X.X.X (unreleased)`, so the
changelog reads: header, a released version, then a section of work that is older than the release
above it but presented as still unreleased.

**2. The tag ships work its release notes do not describe.** `release_data_update.yml` merges the PR
and pushes the version tag off the resulting `master` commit, so the published version contains
everything that had accumulated — but its notes are the single generated bullet, *"updated the data
to `<tag>`"*. The accumulated bullets stay under `X.X.X (unreleased)` and are published again later,
under a version number that did not first ship them. Both halves are wrong: users reading the notes
for the version they installed cannot see what changed, and the eventual entry attributes the work to
the wrong release.

There is a third-order effect worth naming: `update_data.sh` always applies `uv version --bump patch`,
because a data-only release is a patch release. When the unreleased section holds feature work, that
patch bump is publishing it — under a version number that says nothing changed but the data.

## Where this is felt today

The `cut-release` skill treats "a dated section sits above `X.X.X (unreleased)`" as an explicit stop
condition (`.claude/skills/cut-release/SKILL.md:56-62`) rather than silently reordering the file:
reordering would hide the mis-attribution instead of resolving it, and which release the work belongs
to is the maintainer's call. So the situation is currently detected at the next manual release,
by a human, long after the tag went out.

Nothing detects it earlier. No test asserts that `CHANGELOG.rst` sections are in descending order or
that `X.X.X (unreleased)` is the topmost one.

## Options

Not mutually exclusive; (a) is a strict improvement either way.

**(a) Insert below the unreleased section, not above it.** Splice after the `X.X.X (unreleased)` block
rather than after the header. Fixes the ordering. Does *not* fix the mis-attribution — the tag still
covers the unreleased work.

**(b) Refuse to auto-release when the unreleased section is non-empty.** `update_data.sh` (or the
workflow) detects bullets under `X.X.X (unreleased)` and routes to the existing failure path:
`check_data_updates.yml` already labels the PR `automation-failed` and notifies the maintainer. The
data update PR still gets opened and reviewed; only the auto-merge-and-tag step is withheld, so a
release that would publish unreviewed feature work becomes a manual decision. Cheapest correct
behaviour, at the cost of stalling data updates whenever anything is pending.

**(c) Let the data release absorb the unreleased section.** Rename `X.X.X (unreleased)` to the new
version, prepend the data bullet to it, and open a fresh empty `X.X.X (unreleased)` above — which is
exactly what `cut-release` §5 does manually. The notes then describe the tag precisely. The open
question is the bump level: absorbing feature work under a patch bump is wrong, and picking minor
automatically means a script deciding something the `cut-release` skill deliberately stops and asks a
human about.

## Decision: (b), with (a) alongside it

**An upstream data release may not carry unrelated work to PyPI.** When the unreleased section is
non-empty, the auto-merge-and-tag step is withheld and the release becomes a manual decision.

The reasoning is that the automation has no judgement in it and is not meant to acquire any. It is a
shell script and a workflow — no model in the loop — applying an unconditional `uv version --bump
patch`. (c) asks it for exactly the two judgements it cannot make: what bump level accumulated work
deserves, and release notes that describe it. Both belong to the maintainer, which is where the
`cut-release` skill already stops and asks. So (c) is out, and the automated path is restricted to
the case it was written for: a data update with nothing else pending.

The work, then:

- **Detect and withhold.** `update_data.sh` (or `check_data_updates.yml`) checks for bullets under
  `X.X.X (unreleased)` and routes to the existing failure path — label the PR `automation-failed`,
  notify the maintainer. The update PR is still opened and still reviewable; only the merge and the
  tag are withheld. Stalling data updates while work is pending is the accepted cost, and the
  unblocking action is to cut a release.
- **(a) lands too.** The PR still carries a changelog the script rewrote, so the entry must be
  spliced after the `X.X.X (unreleased)` block rather than after the header — otherwise every
  withheld update PR presents an out-of-order file to the human now expected to read it.

## Follow-ups for whoever does this

- Add a test asserting `CHANGELOG.rst` ordering (topmost section is `X.X.X (unreleased)`, dated
  sections descend). This is the check whose absence let the case go unnoticed.
- `.claude/skills/cut-release/SKILL.md` §1 documents this stop condition and §11 already names its own
  removal condition: *"if `update_data.sh` stops inserting above the unreleased section"*. Update the
  skill in the same change.

Adjacent to #501, which covers the *other* unreviewed aspects of the same pipeline (unpinned download,
no data diff, no human in the loop). Same failure surface, different gap: #501 is about not knowing
what the data changed, this is about the release notes not describing what shipped.

Nice6042:
## Summary

Automated data releases could publish unrelated work under a patch version whose notes mentioned only the data update. This change keeps the unreleased section first in the changelog and blocks the merge-and-tag path whenever current `master` contains pending work.

## Changes

- add a fail-closed changelog helper for ordering, pending-work detection, and data-release insertion
- verify the exact update PR, successful head commit, current `master`, and empty unreleased section before merging
- label and notify on blocked automation, with regression tests for the script and workflow wiring

## Related issue

Closes #510

## Validation

- [x] `uv run pytest tests/test_changelog.py tests/test_data_update_workflow.py -q` — 16 passed
- [x] `SKIP=mypy uv run pre-commit run --all-files` — every non-mypy hook passed
- [ ] `make test` — unavailable because GNU Make is not installed; the equivalent pytest command reached 1,235 passes, 8 failures, and 4 errors. The same failures reproduce on `3dca95a` on Windows (C-extension disagreement, Windows pipe behavior, and the Unix-only `resource` module).

## Checklist

The workflow rechecks `master` immediately before merging and pins the PR head SHA. GitHub's merge command has no expected-base-SHA option, so a very small base-branch race remains between that check and the server-side merge.


jannikmi:
Thanks for this @Nice6042 — the design here is the right one, and it's what shipped: read `master`'s changelog immediately before the merge, withhold both the merge and the tag while work is pending, and have `update_data.sh` insert its entry below the unreleased section instead of at the top of the file. Extracting that into `scripts/changelog.py` with its own tests, rather than doing it with `head`/`tail` in shell, is what made the rest reviewable.

I've continued it in #519, which carries your commit as its base and adds the review fixes one commit at a time. The ones worth calling out, since a couple only show up at release time:

- `git tag -a` needs a tagger identity and a GitHub runner has none — the PR would have merged and the release would never have been built.
- `mergeCommit` isn't populated the instant a squash returns, so `--jq .mergeCommit.oid` under `set -e` could abort with the PR already merged.
- `--match-head-commit` pins the PR head but not the base, so the window your `guarded_master` check narrowed was still open. The squash commit's first parent closes it.
- The two notification steps shared one dedup marker, so whichever posted first silenced the other.

The identity check you added to all three steps is what prompted pulling them into a shared composite action — three hand-maintained copies is how the two notices ended up agreeing on one marker.

Closing this one in favour of #519; the credit is recorded in the changelog. Thanks again.

PR #521, opened 2026-08-18
Expose the packaged dataset version at runtime

### LINKED ISSUE #498: Expose the packaged dataset version at runtime
An installed `timezonefinder` cannot say which timezone-boundary-builder dataset it is answering
from, and does not expose its own version either. Verified on `master`:

```python
>>> import timezonefinder
>>> hasattr(timezonefinder, "__version__")
False
```

`DATA_VERSION` exists, but it is a **repo-root file read only by `scripts/configs.py`** — it is not
under `timezonefinder/data/` and so is not in the wheel.

## Why it matters

- **Cache invalidation.** A service caching lookup results has no key to invalidate on when the
  dataset changes. Right now the only proxy is the package version, which also changes for unrelated
  code fixes — and, with the automated data-update pipeline, changes *together with* the data in a
  way callers cannot distinguish.
- **Bug reports.** #393, #356 and #372 are all "wrong timezone at these coordinates". Answering any
  of them starts with "which dataset?", which the reporter currently cannot state without
  reverse-engineering it from the package version.
- **It is already silently broken in published output.** `scripts/benchmark_utils.py` does
  `getattr(tf_instance, "__version__", "Unknown")` on a `TimezoneFinder` *instance* — which has
  `__slots__` and no such attribute — so every generated benchmark report carries
  `timezonefinder_version: "Unknown"` in its provenance block today.

## Proposal

1. `scripts/file_converter.py` writes the dataset version into the data directory it generates.
2. `timezonefinder/configs.py` locates it; `AbstractTimezoneFinder` exposes it as `data_version`.
3. Add `timezonefinder.__version__`.
4. Fix the `getattr` in `scripts/benchmark_utils.py` to read the real value.
5. `MANIFEST.in` / `[tool.setuptools.package-data]` must include the new file, and
   `tests/test_package_contents.py` should assert it is present in the built wheel and sdist —
   that suite already checks for files that must *not* ship; this is the mirror case.

Small, additive, near-zero blast radius. It is also a prerequisite for two larger items: the
compatibility check between a split-out data package and the code (#446), and the diff report on
automated data updates.

MsfPablo:
Closes #498.

## Problem

An installed `timezonefinder` could not state which timezone-boundary-builder release it was answering from. The dataset version lived only in the repo-root `DATA_VERSION` file, which is not packaged, and the package had no `__version__` attribute. As a side effect, every generated benchmark report carried `timezonefinder_version: "Unknown"`, because `scripts/benchmark_utils.py` read a non-existent `__version__` attribute on a `TimezoneFinder` instance (a `__slots__` class with no such attribute).

## Implementation

Follows the 5-step plan in the issue body exactly.

1. **`scripts/file_converter.py` writes the dataset version into the data directory it generates.** After `write_binary_files(...)`, the converter writes `data_version.txt` into the output directory, populated from `read_data_version()` (the existing helper that reads the repo-root `DATA_VERSION`). `update_data.sh` re-stamps both `DATA_VERSION` and `timezonefinder/data/data_version.txt` from `downloaded_tag.txt` after a successful upstream release, so the packaged stamp moves with the data on every update and stays in agreement on a standalone `make parse`.

2. **`timezonefinder/configs.py` locates it; `AbstractTimezoneFinder` exposes it as `data_version`.** `configs.py` declares a single `DATA_VERSION_FILENAME = "data_version.txt"` constant (added to `__all__`) so the build side (file_converter) and the runtime side share one filename and cannot drift. `AbstractTimezoneFinder.data_version` is a `@property` that reads `<data_location>/data_version.txt` and strips it. It is a property, not a slot, so the existing `test_declared_slots_are_assigned` pin is unaffected. `TimezoneFinder` and `TimezoneFinderL` inherit it.

3. **`timezonefinder.__version__` is exposed.** Added in `timezonefinder/__init__.py` via `importlib.metadata.version("timezonefinder")`, with a `PackageNotFoundError` fallback to `"unknown"` for source checkouts without an install. PEP 396.

4. **`scripts/benchmark_utils.py` reads the real value.** `get_system_status()` now imports `__version__` from `timezonefinder` and reports it as `timezonefinder_version`, replacing `getattr(tf_instance, "__version__", "Unknown")` which always fell through to the default. No other field in `get_system_status()` was changed, so `COMPARABILITY_KEYS` and the workload comparison in `test_benchmark_ci_tooling.py` are unaffected.

5. **`MANIFEST.in` / `[tool.setuptools.package-data]` include the new file, and `tests/test_package_contents.py` asserts it is present.** The existing `recursive-include timezonefinder *.py *.npy *.fbs *.txt *.json *.c *.h` in `MANIFEST.in` and the `**/*.txt` glob in `[tool.setuptools.package-data]` already cover `data_version.txt`, so no build-config change is needed (verified by building the wheel and sdist locally and listing their contents). A dedicated `test_packaged_data_version_file_in_distribution` is added to `tests/test_package_contents.py`, parametrised over `DIST_TYPES`, mirroring the existing "must not ship" checks but asserting the stamp is present in both the wheel and sdist.

Additional tests in `tests/test_data_version.py`:
- `test_packaged_data_version_file_matches_repo_root` — the packaged stamp equals `read_data_version()`.
- `test_finder_exposes_packaged_data_version` — `TimezoneFinder().data_version == read_data_version()`.
- `test_package_exposes_version` — `timezonefinder.__version__` is present and not the fallback.

`CHANGELOG.rst` gets a user-facing entry under the unreleased section.

## Verification

- `make test` full suite: 2994 passed (including the new tests, `test_resource_management.py::test_declared_slots_are_assigned`, and `test_package_contents.py` integration builds).
- `ruff check` and `ruff format --check` pass.
- `uv build --python "$(uv python find 3.11)"` produced a wheel and sdist; both contain `timezonefinder/data/data_version.txt` (wheel) / `timezonefinder-8.2.5/timezonefinder/data/data_version.txt` (sdist).
- `rstcheck CHANGELOG.rst` passes.
- pre-commit hooks pass (including `check-manifest`, which required `git add timezonefinder/data/data_version.txt`).
- No `.github/workflows/*.yml` files were modified.

## Scope

Additive and minimal: one new data file, one new `@property`, one new `__version__` lookup, one fixed `getattr`, one added `cp` in `update_data.sh`, and tests + changelog. No public API is removed or renamed; `data_version` is a new read-only property.

## AI disclosure

Authored with assistance from Claude Code. The implementation, tests, and verification were carried out by an automated agent following the issue's 5-step plan; the human operator reviewed and pushed the result. Happy to rework anything that does not fit the project's conventions.


jannikmi:
Thank you for this — the design is right and it is what ships. The stamp written by the converter into the data directory it generates, read back through a `data_version` property, is exactly how the packaged data should state its own release, and finding the `getattr(tf_instance, "__version__", "Unknown")` that had been quietly making every benchmark report useless was a nice catch.

I have continued it in #523, which carries your commit unchanged as its first one, so the authorship stays with you. On top of it are a few review fixes, the main one being that the stamp recorded the repository's pinned `DATA_VERSION` rather than what was actually parsed — so compiling your own GeoJSON produced a directory claiming a boundary release it never came from. The converter now takes the release as `--data-version` (which `update_data.sh` passes from the tag it records at download time, which also removes the `cp` that was repairing the stamp afterwards), and falls back to `"unknown"` for an input nobody attributed.

Closing this one in favour of #523 — please do take a look there, and thanks again for picking up #498.

PR #544, opened 2026-08-21
Address polygon coordinates by offset instead of re-walking the FlatBuffers vtable

### LINKED ISSUE #536: The memory-mapped coordinate accessor costs 4.9 µs per candidate polygon
The memory-mapped coordinate accessor rebuilds a FlatBuffers accessor on every candidate
polygon, which costs **4.9 µs per candidate** — 86× the 57 ns the in-memory mode pays for the
same bytes, and 47% of an ambiguous `timezone_at` query.

Found while profiling for #497; see the [measurement](https://github.com/jannikmi/timezonefinder/issues/497#issuecomment-5339647913)
for the full stage breakdown and `prototypes/query_stage_profile.py` to reproduce.

## What is happening

`FileCoordAccessor.__getitem__` calls `read_polygon_array_from_binary`, which does
`poly_collection.Polygons(idx)` → `poly.CoordsAsNumpy()` → reshape on every access. Measured on
one *warm* polygon, 20,000 repetitions, so this is not page-fault cost:

| step | ns |
|---|---:|
| `PolygonCollection.Polygons(idx)` (FlatBuffers vtable walking, in Python) | 2,129 |
| `CoordsAsNumpy()` | 2,687 |
| `reshape_to_polygon_coords` | 145 |
| **`coords_of`, total** | **5,012** |
| the same via `MemoryCoordAccessor` (`in_memory=True`) | 84 |

Consequence per stratum, on this data (numba backend, local):

| | mapped | in memory |
|---|---:|---:|
| ambiguous-shortcut query | 13.3 µs | 9.0 µs |
| one small-polygon PIP (fetch + kernel) | 5.15 µs | 0.30 µs |

So on the default mode, a candidate polygon under ~3,000 vertices costs more to *open* than to
test.

## Why this is worth fixing rather than answering with "use `in_memory=True`"

`in_memory=True` reads the whole ~63 MB coordinate file into the heap, per instance — and the
documented pattern for concurrent workloads is one instance per thread. The mapped mode exists
for the constrained-CPU/memory container deployments this package targets, and it is the
default. It is currently paying 5 µs per candidate for a saving it does not have to make.

## The shape a fix could take

The per-polygon views can be built **once, over the mmap**, instead of per access — no
coordinate bytes are copied, so the mapped memory profile is unchanged. Measured on the current
data (1,322 boundary polygons):

* building every view up front: **8 ms**, once, at construction
* Python object overhead: ~165 KiB total
* fetch afterwards: **30 ns** (a list index), against 4,894 ns

That is not a proposal to merge as written — two things need working out first:

1. **Initialisation cost.** 8 ms is real and is paid per instance, i.e. per thread under the
   documented pattern. It may be better to precompute the (offset, length) pairs and build the
   view lazily per polygon, or to store the offsets in the data file rather than deriving them.
   Whatever is chosen, `benchmarks/test_initialization.py` and `make memory` are the two places
   it has to be shown not to regress — the point of the mapped mode is its footprint.
2. **Holding 1,322 live views pins the mapping.** `FileCoordAccessor.cleanup` already documents
   that `mmap.close()` is refused while views handed out by `__getitem__` are alive, and
   deliberately drops its own references so the unmap happens once the last view goes. Caching
   the views inside the accessor makes the accessor itself the holder, so that reasoning has to
   be redone rather than inherited.

Ranking note: this sits under the same lever as #364 (a native candidate loop crosses into C
once per query and would not touch a Python-side accessor at all), and it is measured against
the same query the batch API (#499) is about. Which of the three is worth doing first is a
sequencing question for #506, not for this issue.

---

## Update, 2026-08-21 — what is still open, and what has been decided elsewhere

This is now the **largest measured lever** in the improvement register: 47 % of an ambiguous query
and **~37 % of a uniformly random workload**, paid by exactly the constrained-memory deployments
the mapped mode exists for. What it removes is machine-independent — one accessor rebuilt in
Python per candidate polygon, at 1.13 candidates per ambiguous query.

It is also the **denominator for the geometry work**: landing it roughly halves what a candidate
costs, so anything ranked on the *time* in the candidate loop (#364, #513) has to be priced after
it, not before. `potential-improvements.md` holds that sequencing; the roadmap issue #506 that the
note at the bottom of this body points at has been closed and replaced by that file.

### Why this is a decision rather than a ready-to-take item

The fix **caches something**, and both consequences of that are choices only the maintainer can
make. Restating the two open questions above in the form they now need answering:

**1. Where the build cost lands.** Building every view up front is 8 ms per instance, and the
documented concurrency pattern is one instance per thread — so it is 8 ms × threads, on a path
that is itself a tracked benchmark (`docs/benchmark_results_initialization.rst`).

*Half of this is now settled by precedent.* On 2026-08-20 the same question was decided for
`zone_positions.npy` (a 1,018-byte array re-loaded on every `certain_timezone_at` call): **cache
lazily on first use, never eagerly at construction**, because data serving an optional path should
not charge every user at construction time. That rule applies here with more force, not less — 8 ms
is four orders of magnitude more than the case it was decided on. So the up-front build is out, and
what remains is lazy-per-polygon: either building a view on first access to that polygon, or
precomputing only the (offset, length) pairs, or storing the offsets in the data file so nothing has
to be derived at all.

**2. How much of the mapping the cache is allowed to pin.** This is the half the `zone_positions`
precedent does **not** settle, and it is the real decision. There the file was a kilobyte, so
holding it open cost nothing; here the mapping is ~63 MB, and a cache of live views makes the
accessor the holder that keeps it mapped — which is a memory cost in the mode that exists to avoid
one. `FileCoordAccessor.cleanup` currently drops its own references precisely so the unmap happens
when the last handed-out view dies; caching inside the accessor inverts that and the reasoning has
to be redone rather than inherited.

Three shapes, to be chosen between:

| | build cost | pinned | notes |
|---|---|---|---|
| **A — bounded lazy cache** | none at construction | bounded by the cache size | needs a size or eviction policy, which is the part to decide. Most candidates are re-tested across nearby queries, so even a small cache should hit |
| **B — unbounded lazy cache** | none at construction | grows toward the whole mapping | simplest and fastest; the working set converges on what `in_memory=False` exists to avoid |
| **C — no cache; a native candidate loop** | none | none | removes the question instead of answering it, by never building a Python-side accessor — this is #364's native loop, at L size |

Whichever is chosen, `benchmarks/test_initialization.py` and `make memory` are where it has to be
shown not to regress, and the register's rule applies to the result: state the benefit as the count
removed and as a share of a mixed workload, not as absolute nanoseconds off one machine.

### One thing that turned out **not** to depend on this

#301 (sort shortcut polygons by overlap area) was going to be re-measured after this landed, on the
grounds that this item changes what a candidate costs. That was the wrong instrument: what ordering
controls is how many candidates are **tested**, which is a count and does not depend on their cost.
Enumerated over the packaged index, the whole headroom is 366 tests — 2.90 %, on 259 of 41,162
cells — so #301 is refused on its own terms and needs nothing from this issue.

---

## Correction, 2026-08-21 — there may be no decision here at all

**The section above frames this as a choice between cache lifetimes, and that framing assumes the cache holds *views*.** It does not have to. Found while scoping #364 against a free-threaded interpreter.

The expensive part is not the view — it is **re-walking the FlatBuffers vtable in Python to find where polygon *i*'s bytes are**. Cache the `(byte offset, length)` pair per polygon instead:

```python
np.frombuffer(buf, "<i4", count=LEN[i], offset=OFF[i]).reshape(2, -1)
```

That is **~10 KiB as `uint32`** for 1,322 polygons. Integers pin nothing, so:

- **the cache-lifetime-versus-pinning trade does not arise** — the accessor never becomes a holder of live views, and `FileCoordAccessor.cleanup`'s existing reasoning is inherited unchanged rather than redone
- **the 8 ms up-front build does not arise either** — deriving the table was measured at ~0.5 ms, and it need not be eager in any case

### Measured

| | ns |
|---|---:|
| `coords_of` today, mapped path | **5,646** |
| `np.frombuffer` + `reshape` (what the table reduces it to) | **568** |
| removed per candidate | **~5,078 (9.9x)** |

That is essentially all of what this issue is about, **with no C, no data-format change and no cache policy**. Absolute nanoseconds are one machine's; the ratio and the count are not.

### What is not yet verified

That the table can be **derived correctly for every polygon** and that the result is byte-identical to today's reader across the whole dataset. A scratch prototype produced it in 0.52 ms with a vectorised walk (assuming the layout the current writer produces) and 6.9 ms with a layout-agnostic Python loop, and matched `coords_of` on 12 random polygons — but that is a prototype, not a reviewed implementation. **Checking it across all 1,322 polygons is the first thing an implementing pass should do**, and it is cheap.

### Consequences if it holds

- this stops being a decision and becomes **~40 lines of pure Python**
- it becomes the prerequisite **#364's native candidate loop** needs, since handing C a base pointer plus an offset table is far simpler than implementing FlatBuffers traversal in C
- and it sharpens #364's own justification into a question: after ~5.1 µs of ~9.3 µs per candidate is removed here, what remains of the candidate loop may sit inside the benchmark suite's own noise floor, in which case a native loop cannot be justified on speed at all

The **`Ranking note`** at the end of the original body is also out of date: #506 is closed and replaced by `potential-improvements.md`, and #301 has been closed as not planned — its ordering question was refused on a count (2.90 % of point-in-polygon tests, on 0.63 % of cells) that does not depend on what a candidate costs, so it never needed to wait for this issue.

jannikmi:
> **Stacked on #541**, which carries the register pass that identified this item — this PR's base is that branch, so the diff here is the implementation alone. GitHub retargets it to `master` when #541 merges.

Closes #536.

On the memory-mapped path (`in_memory=False`, the default), fetching a candidate polygon's coordinates rebuilt a FlatBuffers accessor in Python on every lookup — `PolygonCollection.Polygons(idx)` plus `CoordsAsNumpy`, **~4.9 µs per candidate**, 47 % of an ambiguous query, paid by exactly the constrained-memory deployments the mapped mode exists for.

Each accessor now resolves every polygon's `(byte offset, length)` once and afterwards slices coordinates straight out of the buffer.

| clang / mapped | before | after |
|---|---:|---:|
| `coords_of`, per candidate | 4,894 ns | 834 ns |
| ambiguous query | 14,016 ns | 10,718 ns |
| `on_land` query | 8,889 ns | 6,900 ns |
| cost of mapped mode vs `in_memory=True` | +46 % | +15 % |

Coordinates, results, the public API and construction time are unchanged. Queries answered from a single-zone shortcut cell never fetch coordinates and are unaffected.

### Two things that are not obvious from the diff

**Integers, not arrays.** Caching the polygon arrays would be shorter and just as fast, but each cached array is a zero-copy export of the memory map, so the cache would pin the mapping for the accessor's lifetime — in the mode that exists so the data need not be held in memory. Not hypothetical: `mmap.close()` refusing while an export is alive is the `BufferError` fixed in 8.3.0. Two `uint32` vectors pin nothing and cost ~10 KiB. `tests/test_coord_offset_table.py` asserts this by unmapping — with only the table alive `mmap.close()` succeeds, with one fetched view alive it raises.

**Whole-array arithmetic, not a loop over `Polygons(i)`.** The obvious derivation costs **~6.2 ms** for the packaged boundaries — materially the same as caching views would have — and a per-construction cost of that size is what would make *when* to build the table a decision, one paid again by every thread. Walking the vtables with numpy costs **~0.1 ms** against a ~400 ms mapped construction, so it is simply built eagerly and there is nothing to weigh.

### How it is checked

The fast path now assumes what the reader guarantees — that a table sits where its reference says, that vtables are aligned, that the coords field is present — and a violated assumption is not self-announcing: an offset landing one word early still yields plausible `int32` coordinates, i.e. a plausible wrong timezone.

So `scripts.data_integrity.validate_coordinate_offset_table` compares the table against `read_polygon_array_from_binary` for **every polygon in both coordinate files**, run by the converter over what it just wrote and by the test suite over what is committed — the pattern the hole-reference checks next to it already use. Exhaustive because it is off the init path. Verified before any of this was written: byte-identical for all 1,322 boundary polygons and 27 hole rings, and identical `timezone_at` answers over all 30,000 committed fixture points.

### Re-measurement

The register requires the PR that moves the critical path to re-run `prototypes/query_stage_profile.py` and update every share it moves, so `FINDINGS` is re-measured on both backends and both access modes, and re-anchored. Replaced wholesale rather than edited row by row: the two runs are a different interpreter and a different day apart, so stages that did not change still move by the machine's own 3–9 %.

Two conclusions change beyond finding 5 being closed — the fetch no longer dwarfs the ~650 ns FFI marshalling, and the fetch-versus-kernel crossover drops from around ten thousand vertices to roughly 1,400, which makes the medium polygon stratum kernel-bound.

The re-measurement turned up its own follow-up, recorded as **PERF-4** rather than taken here: `np.frombuffer` re-acquires the `mmap`'s buffer per fetch, and slicing one whole-file view instead measures 415 ns against 788 ns — but that reintroduces exactly the accessor-lifetime pinning this design avoids, so it needs a decision.

Incidental: the documented command for running the profiler against the C extension had never worked — `--isolated` drops the default dependency groups, so it could not import `tests/auxiliaries.py`.

### Register

GH-536's entry and ranking row are deleted, and the five cross-references that treated it as pending are corrected rather than dropped — including the two recorded decisions that cited its pinning trade as live.

### Testing

`make testall` green (3,170 passed, 1 skipped), plus the coordinate, acceleration-path, resource-management and main suites re-run in the bare `--no-group numba` environment, since a local run binds numba at import and would otherwise never exercise the C extension.


jannikmi:
The `init_rss` flag was real, and it is fixed in fcd032c. Thanks, CI — the local measurements did not catch it.

**Cause.** The header words the offset table is derived from are ~5 KiB in total, but they sit next to the polygons they describe, so they are spread over the whole file — **788 distinct pages of the packaged boundaries, across 60 MiB**. Gathering them through the mapping faults every one of those pages in, and the kernel's readahead multiplies that. `steady_rss` was unchanged (+2.2 %) because a workload touches those pages anyway; what regressed was specifically a finder that has answered *no* queries yet — the case the mapped mode exists for.

**Fix.** The walk now takes its scattered words through a reader, of which there are two: whole-buffer views where the bytes are already in memory (the in-memory accessor, the integrity check), and `seek`+`read` on the coordinate file for the mapped accessor. Those bytes land in the page cache — shared, reclaimable, nobody's resident set — and ~5 KiB is copied to the heap.

Measured with this repository's own probe, min of 3 runs:

| | base | as pushed | fixed |
|---|---:|---:|---:|
| `init_rss` | 12.89 MiB | 21.16 MiB | **13.53 MiB** |
| `init_heap` | 4.78 MiB | 4.79 MiB | 4.54 MiB |
| `steady_rss` | 45.61 MiB | 47.97 MiB | 45.73 MiB |

Construction is unchanged at ~387 ms — the file reads cost ~4 ms against a 389 ± 8.9 ms baseline. The file is opened **unbuffered**: through a buffered object each seek discards a read-ahead buffer the next seek refills, which measures 6× slower over the same reads. Nothing else reads through that object, so a later tidy-up back to `open(path, "rb")` would look harmless — a test pins it.

Query timings are untouched; nothing on the lookup path changed.

---

---

**Separately, worth flagging: the comparison above cannot see what this PR is for.**

Not because the file-based mode is unmeasured - `TIMEZONE_AT_CASES` carries a `-file_based` case for all four fixtures, and they run under `make benchmarks` and feed the report pages. It is that none of them carries the `benchmark_core` mark, and the PR-to-merge-base job runs `pytest benchmarks -m benchmark_core`. The core set is `in_memory` only, deliberately: the file-based mode has its own I/O noise profile and would widen the one shared alert threshold for everything.

The memory report has no such restriction - `scripts/measure_memory.py` carries its own `CONFIGS`, and `TimezoneFinder[file_based]` is one of the three - which is exactly why CI could see this change's cost and not its benefit. An ambiguous mapped query going 14.0 to 10.7 us shows up above as three unchanged in-memory rows.

That is a coverage gap rather than something this PR should fix inline: the reason for keeping the mode out of the core set is sound, and the fix is a separate alert threshold for it rather than more rows on the shared one. Happy to open it as its own item.

jannikmi:
Confirming for the record: the table is built **eagerly**, in `FileCoordAccessor.__init__` / `MemoryCoordAccessor.__init__`, and there is no lazy path anywhere on the change. bcd78f5 does not alter that - it replaces the *reason*, which was wrong in a way that would have invited a lazy rewrite later.

The justification as written rested on the derivation being cheap ("at 0.1 ms there is nothing to weigh"). That is an argument that expires: it says defer as soon as the cost grows. The reason that survives is categorical.

**Polygon coordinates are not optional-path data.** A `TimezoneFinder` exists to test points against polygons - every query not answered outright by a unique-zone shortcut cell reaches the accessor, and there is no population of callers who never do. Deferring the table would move a certain cost to the first query rather than avoid one. It would also buy that with two things worth more than the milliseconds:

- an `is None` branch per fetch, on the hot path this PR exists to shorten;
- **a write to `self` from a lookup** - and that is the expensive half. A shared instance is currently safe for concurrent reads precisely because every attribute is assigned in `__init__` or `cleanup()` and nothing on the lookup path mutates state (GH-364's finding (c)). A lazy cache would be the first thing to break that, silently and only under load.

This is the *opposite* case to the lazy rule that governs `zone_positions`, which serves `certain_timezone_at` and `get_geometry` - an optional query mode the `timezone_at` majority never calls. The register's recorded decision now states both sides and why they are not symmetric, so the next pass does not read one as a precedent for the other. The derivation cost decides *how* the table is derived, never *whether* to defer it.

`test_a_lookup_mutates_no_accessor_state` pins the invariant on both accessors - by identity rather than equality, so a rebuilt-but-equal array still counts as a write - and was verified against a deliberately lazy mutation of `FileCoordAccessor`, which it fails.

`make testall` green (3,175 passed, 1 skipped).

jannikmi:
Two pieces of bookkeeping this PR owed and had not done, now in 448b926.

## The report pages are regenerated

`docs/benchmark_results_*.rst` are committed *measurements*, so a change that moves the critical path leaves them describing a tree that no longer exists - and nothing fails, because the pages still build and render. `make reports` re-measures and re-renders all four. `docs/data_report.rst` is unchanged, as it should be: the packaged data did not move.

| | before | after |
|---|---:|---:|
| ambiguous-shortcut, file-based | 37.0 ms | **27.0 ms** |
| on-land, file-based | 20.4 ms | **16.0 ms** |
| random, file-based | 11.7 ms | **8.97 ms** |
| `timezone_at_land`, file-based | 21.8 ms | **17.2 ms** |
| heap, `TimezoneFinder[file_based]` | 4.77 MiB | **4.54 MiB** |

The heap drop is the `PolygonCollection` no longer being retained.

**Read the rest with the machine caveat**, because `make reports` re-measures *everything* on the current machine and the previous pages were taken on another day. The in-memory timing rows move +2-9 % and every initialization row +4-5 %, and neither is this change: `TimezoneFinderL`'s construction moved +17 ms, and it holds no `PolygonArray` at all. So the ~4 ms the offset table costs at construction sits inside a noise band that moved every configuration, including the ones this change cannot touch. Same machine class as the pages they replace (Darwin arm64, Python 3.14.2, Numba path), not comparable with the CI figures above.

`make docs` builds with no new warnings.

## The GH-536 handle is fully retired

Deleting a shipped register entry is not the whole job. The entry and its ranking row went with the implementation, but **eleven references still named `GH-536`** - the dependency graph, *Sequencing*, two other entries' preconditions, three recorded decisions. Each was a handle resolving to nothing, and each read as live status in a file that is explicitly a to-do list rather than a history, so the next pass would either hunt for an entry that is not there or price work against a blocker that no longer exists.

Every one now describes the thing instead of naming the closed issue - what a candidate polygon costs, rather than "after GH-536". Recorded decisions are kept, so they were rewritten this way rather than deleted. `grep GH-536 potential-improvements.md` is now empty.

`CLAUDE.md` states the obligation, since only the entry and its row were written down before.

PR #550, opened 2026-08-23
Measure the tzfpy comparison instead of asserting it

jannikmi:
## Summary

`docs/alternatives.rst` weighed this package against `tzfpy` on rows that had never been measured together — the published figures came from different machines, different acceleration paths and different query workloads. That is exactly the comparison this project refuses to make about its own numbers, so it had no business making it about somebody else's.

This adds a benchmark suite that runs both packages over the **same committed query fixtures, in the same process, on the same machine**, publishes the result as a generated page, and corrects the comparison page against it.

**The measurement confirms one row and refutes another.**

`tzfpy` is still clearly the faster one per lookup, and the gap has not closed:

| Query points | tzfpy | TimezoneFinder (in-memory) | vs tzfpy |
|---|---|---|---|
| random (representative mix) | 313 ns | 1.72 µs | **5.47x slower** |
| on-land | 357 ns | 3.40 µs | 9.51x slower |
| unique-shortcut | 273 ns | 823 ns | 3.02x slower |
| ambiguous-shortcut | 626 ns | 10.2 µs | 16.4x slower |

If anything the old wording — "both packages are in the same order of magnitude" — was generous to this one: the class that falls through to a full point-in-polygon test is over an order of magnitude apart. The shape is exactly what the design difference predicts, and it is the price of the accuracy this package exists for.

What the page **got wrong** is startup. It claimed `tzfpy` needs "None (immediate)". It does not: it imports in about a millisecond and then deserialises its index lazily inside the *first query*, which costs about what this package spends importing NumPy and H3 and reading its own index. Net of a bare interpreter, time to a first answer is ~95 ms for `tzfpy`, ~100–105 ms for `TimezoneFinder`, ~87 ms for `TimezoneFinderL`. Startup is therefore no longer listed as a reason to choose either package.

## Changes

- **`benchmarks/test_comparison.py`** — per-query time for `TimezoneFinder`, `TimezoneFinderL` and `tzfpy` over the four committed point fixtures, plus time to a first answer in a fresh interpreter with a bare-interpreter baseline row. Each library is called through its own API inside its own loop: normalising `timezone_at(lng=, lat=)` and `get_tz(lng, lat)` behind one wrapper would cost a Python frame against per-query times of a few hundred nanoseconds, which is the same order as the thing being measured. `TimezoneFinderL` is measured alongside as this package's own version of the same accuracy-for-speed bargain.
- **A new generated page**, `docs/benchmark_results_comparison.rst`, rendered by `make reports` and stamped with the `tzfpy` version it measured — that package releases outside this repository, so a page of ratios that does not name its denominator goes stale silently.
- **`docs/alternatives.rst`** — speed row now cites the measurement, startup row corrected, the "never benchmarked under one harness" note replaced with what the harness found and what it deliberately does not settle (`tzfpy`'s memory footprint; accuracy, since counting disagreements cannot say which answer is right). The decision table drops "Initialization time". Also fixes a stale "H3 resolution 3 (~41k cells)" left over from the resolution change in the base branch.
- **`compare` dependency group** for `tzfpy`, asked for explicitly by `make benchmarks` and `tox -e benchmarks`, and deliberately absent from the CI-tracked environment (`uv sync --group test`), which must stay exactly what a plain `pip install timezonefinder` gives you.
- **`docs/benchmarking_methodology.rst`** gains a section on cross-package comparison; `CONTRIBUTING.md`, `CLAUDE.md` and `docs/7_performance.rst` updated.

**Not tracked on the trend chart.** No benchmark here carries `benchmark_core`. The number is a ratio against a package this repository does not control the release schedule of, so a step in that chart would as often mean "they shipped" as "we changed", and nothing on the chart could tell a reader which.

## One change that deserves review

Measuring startup exposed that **the committed reports described the developer's environment rather than the plain install.** `make install` syncs `--all-groups`, and `timezonefinder/utils.py` binds the point-in-polygon backend at import time preferring numba whenever it is importable — so every local `make reports` silently published numba numbers. For the lookup timings that was a modest distortion; for a time-to-first-answer comparison it is fatal, because importing numba alone costs a few hundred milliseconds, an order of magnitude more than the difference being measured.

So `make benchmarks` and `make memory` now measure in an `--isolated` environment (`BENCHMARK_ENV` in the `Makefile`) and assert the acceleration path before recording anything, the same guard the benchmark workflow already applies. **All five report pages are regenerated from one run of it**, which is why three otherwise-unrelated pages show diffs: they now carry "This is the configuration continuous integration tracks" instead of a warning that they are not comparable to it.

This is a deliberate change to a shared artifact and is easy to revert — one line in the `Makefile` — if you would rather keep publishing numba figures.

The `-ci` targets are untouched: CI syncs its own environment and asserts the same value before measuring.

## Related issue

None.

## Validation

Measured on **Darwin arm64, Python 3.14.2 (free-threaded), C extension without numba** — not comparable with CI's runners.

- [x] `make test` — 1354 passed, 1799 deselected
- [x] Other relevant checks:
  - `make hook` — all hooks pass
  - `make docs` — build succeeded; the one warning (`docs/data_format.rst:146`) is pre-existing and untouched here
  - `make benchmarks` / `make memory` / re-render — run twice, before and after the rebase onto `130d645`; the head-to-head ratios agreed to within ~1% across runs
  - the acceleration guard was verified to fail (exit 1) when numba is present
  - the skip path was verified by blocking the `tzfpy` import: 5 skipped, node id set unchanged
- Slow tests were not run. This change touches benchmark tooling, reporting code and docs — none of `polygon_array.py`, `coord_accessors.py`, shortcut generation, the data converter, `timezonefinder.py` or the `utils*` modules — which is the case `CLAUDE.md` says to skip them for.

## Checklist

### Known risks and limitations

- **One machine.** The ratios survive a change of hardware far better than the absolute numbers, but neither is a promise, and the page says so.
- **The comparison is against a moving target.** `tzfpy` 1.3.3 is what was measured; the version is recorded in the report, and re-running `make reports` is what keeps it current.
- **`tzfpy` ships abi3 wheels for CPython 3.10+** on every platform this project builds for. On one it does not cover, `uv sync --all-groups` would need a Rust toolchain — `--no-group compare` is the escape, and only the comparison is lost.
- **Accuracy is not scored.** The two packages do disagree on a small fraction of points, but settling who is right needs ground truth that neither package carries, so the page states the design difference instead.

---

Stacked on #545 — targets `register/shortcut-layout-measured`, and the report figures were re-measured after rebasing onto its current tip.


jannikmi:
@ringsaturn have a look

ringsaturn:
Thanks for taking the time to put together a proper comparison. The initialization cost is already mentioned in tzfpy's README as "lazy init," but I agree that it could be clearer. I may add a short warm-up example such as `_ = get_tz(0, 0)`, which forces tzfpy to load the required data before serving requests.

Regarding accuracy, tzfpy simplifies the boundaries with a maximum displacement of roughly 111 meters. That trade-off made sense for my original use case, which was a weather data API where coordinates are rarely sensitive to such a small difference near timezone borders.

TimezoneFinder is already fast enough for most applications. The main issue I encountered was the long-tail latency of queries near borders, which can be seen in this [snapshot](https://github.com/ringsaturn/tz-benchmark/blob/main/snapshot/2026-07-26-8d0fed77a8efb102ea3e3848781b5a000bbfb548/benchmark_result_python.txt). That was one of the reasons I created tzfpy. For CPU-intensive Python backend services, a small number of slow queries can have a noticeable impact on the overall p99 latency.

For benchmarking, GitHub Actions runners tend to have noisy and heavily contended CPUs, so I would treat their absolute numbers with caution. I have been planning to run the benchmarks periodically, perhaps monthly, on a fixed AWS instance. I have not had enough time to set that up yet, though.

There is also some ongoing work in the tzf project to move to a custom binary format. This should reduce both memory usage and initialization time. I hope to release it sometime in Q3 or Q4 2026, depending on how much time I can allocate to it.

Overall, the results and the explanation of the design trade-off look reasonable to me. Thanks again for including tzfpy in the same benchmark harness.

Issue #542, opened 2026-08-20
Establish what coordinate precision is worth: what competitors offer, and what users need


jannikmi:
`timezonefinder` stores coordinates as int32 scaled by 1e7, i.e. **~1.1 cm** resolution, and `docs/data_format.rst` advertises that figure. `CLAUDE.md` calls it "a deliberate ceiling, not a bug", and the package's stated selling point is accuracy at timezone borders — no geometry simplification.

That precision is also the single largest lever on the packaged data's size, and #449 cannot be decided without it. This issue exists to establish the facts, so that #449 becomes a decision about encoding alone.

## Why this blocks #449

#449 measured three encodings of the boundary coordinates:

| encoding | payload | decode, median polygon | decode, largest polygon |
|---|---:|---:|---:|
| today (int32, per-axis) | 63.4 MB | 5.6 µs | 5.6 µs |
| delta + zigzag + varint, **lossless at 1e-7** | 35.3 MB | ~34 µs | ~8.5 ms |
| int16 delta + per-delta escape, **at 1e-6** | ~32 MB | ~4.7 µs | ~828 µs |

The rows are not comparable: the second keeps today's precision, the third spends it. ~3 MB separates them, and dropping to 1e-6 (~11 cm) would change a documented guarantee for that 3 MB. Nobody can weigh 3 MB against 11 cm without knowing what the precision is *for*, which is what this issue is meant to answer.

Note also that precision alone buys almost nothing: quantizing to 1e-5 while keeping the current int32 layout moves deflate only 52.5 → 49.9 MB (−5 %). It only pays combined with delta+varint encoding. So this is not "should we reduce precision to save space" — it is "is the 1.1 cm promise load-bearing, and would anyone notice if it became 11 cm".

## What to establish

**1. What precision do comparable libraries actually offer?** In particular [`tzfpy`](https://github.com/ringsaturn/tzfpy) / [`tzf-rs`](https://github.com/ringsaturn/tzf-rs), which is the closest competitor on the same trade-off axis and is explicit about using a simplified ("lite") dataset by default. Establish, with evidence rather than from README claims:
   - the coordinate precision of its packaged data, and whether the default build is simplified
   - the resulting disagreement rate against this package over a large uniform sample and, separately, over points sampled *near borders*, which is where the two designs actually differ
   - its package size for that precision, so the size/accuracy point it occupies is comparable to ours

   A prototype under `prototypes/` is the right home for the comparison, so it can be re-run.

[...]

Issue #504, opened 2026-08-13
CLI: read coordinates from stdin instead of one pair per process


jannikmi:
`timezonefinder <lng> <lat>` builds a finder, answers one point, and exits. Looking up a thousand
coordinates means a thousand processes, each paying full initialisation to answer one query.

Proposal: a streaming mode reading `lng,lat` per line from stdin and writing one result per line.

```console
$ cat points.csv | timezonefinder --stdin
Europe/Berlin
America/New_York
```

The instance is constructed once, so initialisation amortises across the whole input, and the CLI
becomes usable in a shell pipeline.

Fits naturally with the existing contract: `tests/cli_test.py` already pins that nothing but the
result is written to stdout, precisely so the output can be piped. This extends that property to
more than one line.

Notes for whoever picks this up:

- keep the existing single-pair form working unchanged
- decide what a malformed or out-of-range line does — skip with a warning on stderr, or abort;
  aborting on line 999 of 1000 and discarding the rest is hostile
- `-f/--function` should apply to the stream as a whole
- the natural companion is the batch lookup API, if that lands — the CLI would be its first consumer

`timezonefinder/command_line.py`, `tests/cli_test.py`.

kimberly-WY:
One thing worth trying: setting the timeout at the transport layer rather than the request level — avoids the race condition.

Issue #499, opened 2026-08-13
Batch / array lookup API: answer many coordinates per call


jannikmi:
`CLAUDE.md` names the primary users as *"latency-sensitive services doing high-volume, possibly
concurrent lookups"*. The API answers **one point per call**, so every one of those users writes a
Python loop — which is exactly what `examples/batch_processing.py` recommends as best practice.

## Why this is not just ergonomics

From the measurements in #477 (numba backend, macOS arm64):

| query | total |
|---|---|
| `TimezoneFinder`, unique-zone hit (Berlin) | **0.94 µs** |
| `TimezoneFinder`, ocean hit | 0.96 µs |
| `TimezoneFinder`, NYC | 14.9 µs |
| `TimezoneFinder`, near a border (Basel) | 28.9 µs |

The first two touch **no geometry at all** — the hybrid shortcut resolves them to a zone id
directly, and ~30,651 of the 41,162 index entries are that case. So ~1 µs is being spent almost
entirely on per-call overhead: `validate_coordinates`, `h3.latlng_to_cell`, `shortcut_mapping.get`
(68 ns of it, per #477) and `zone_name_from_id`. That is a **~1 M lookups/s ceiling made of nothing
but interpreter and call overhead**, and it is precisely what a batch path amortises.

`TimezoneFinderL` is *only* that path, so it is affected across the board.

## Sketch

```python
tf.timezone_at_many(lngs, lats)   # -> list[str | None] or np.ndarray of ids
```

- vectorise validation and the int scaling (`coord2int`) over the input arrays
- resolve every point's shortcut in one pass; answer unique-zone cells with a numpy gather
- fall through to the polygon loop only for the residual ambiguous points, **grouped by cell** so a
  candidate polygon is fetched once per group instead of once per point
- offer an id-returning variant, so a caller doing 10^6 lookups is not paying for 10^6 string
  lookups it will immediately map back to something else

## Relationship to other issues

- **#477 is the enabler.** The dict cannot be vectorised; flat arrays can. Note the scalar-path
  regression that makes #477 a hard call *does not apply here* — a batch path has no scalar lookup
  to slow down. That may be the strongest argument for landing #477's flat layout at all.
- **Blocked on profiling** (see the profiling issue): the claim above is that overhead dominates the
  unique-zone path. #477's totals support it strongly but do not break the 0.94 µs down by stage.

## Open design questions

[...]

Issue #449, opened 2026-07-31
rethink the polygon binary storage encoding


jannikmi:
## What the format is today

`boundaries/coordinates.fbs` is 63,423,680 B, of which **63,402,504 B is payload and only 21,176 B (0.03%) is FlatBuffers overhead**. It's 1,322 polygons / 7,925,313 vertices stored as interleaved int32 pairs — `[x0,y0,x1,y1,…]` — at 1e-7 degree. So there is nothing to reclaim by swapping the container; every available byte is in *how a vertex is encoded*. That's why this is a data-layout project, not a packaging one.

Size distribution matters for everything below: median polygon is **324 vertices**, max is **192,960**, and the **192 polygons over 10k vertices hold 88% of all vertices**.

## Three independent transforms

**(a) AoS → SoA.** Store `[x0…xN][y0…yN]` per polygon instead of interleaving. Compression gain is small on its own — deflate 52,545,858 → 51,222,930 (−2.5%) — but it's free and it fixes a real inefficiency. Today `coords_of()` returns a `(2,N)` view reshaped with `order="F"`, so `c[0]` has **stride 8 and is not contiguous**. `pt_in_poly_clang` then calls `np.ascontiguousarray` on both rows, which copies on every PIP test:

| | `coords_of()` | `+ ascontiguousarray` (clang path) |
|---|---|---|
| median (326 verts) | 5.58 µs | 0.8 µs |
| largest (192,960 verts) | 5.63 µs | **122.6 µs** |

SoA makes both rows natively contiguous and that copy disappears. (Numba users don't pay it — `pt_in_poly_python` handles strides — so this only helps the clang-without-numba path.)

**(b) Delta encoding.** Consecutive boundary vertices are close, so deltas are small. At full 1e-7 precision the zigzag delta needs a mean of **2.23 bytes** per component vs 4.00 fixed:

```
1 byte:  2.46%      3 bytes: 25.04%
2 bytes: 72.38%     4+:       0.13%
```

**(c) Varint.** Cashing in (b) requires a variable-width code. Delta + zigzag + LEB128 over SoA:

| precision | on-disk payload | vs raw |
|---|---|---|
| 1e-7 (today's precision, lossless) | 35,320,271 B | 0.557 |
| 1e-6 | 29,043,537 B | 0.458 |
| 1e-5 | 20,613,104 B | 0.325 |

Note the top row: **−44% with no precision loss at all.** Precision is a separate, optional axis on top.

[...]

PR #552, opened 2026-08-23
Answer many coordinates per call: timezone_ids_at / timezone_names_at

jannikmi:
## What

Adds the batch lookups from the register's top-ranked item, GH-499, on `TimezoneFinder`, `TimezoneFinderL` and as global functions:

```python
zone_ids = tf.timezone_ids_at(lngs=[13.358, 2.3522], lats=[52.5061, 48.8566])  # int32 ndarray
names    = tf.timezone_names_at(lngs=[13.358, 2.3522], lats=[52.5061, 48.8566])
```

- **Ids are primary.** A caller doing millions of lookups should not pay for millions of string lookups it maps straight back to something else; `timezone_names_at` is the convenience on top.
- **One array per axis**, any 1-D array-like; a C-contiguous `float64` numpy array is used without copying. An `(N, 2)` array is rejected rather than read by column position.
- **`NO_ZONE_ID` (`-1`)** is exported from the package and stands where the scalar methods answer `None`.
- **`on_invalid="raise" | "skip"`**, defaulting to `"raise"` to match the scalar methods. `"skip"` answers the rest of the batch rather than discarding the work already done for it — always one answer per input coordinate, in input order.
- **Scope: `timezone_at` only.** `timezone_at_land` and `certain_timezone_at` do not get a batch form here; the reason is under *Next*.

Internally the candidate loop that resolves a point in a multi-zone cell now exists once, behind `_zone_id_in_ambiguous_cell`, which the scalar `timezone_at` calls too, and `ShortcutIndex` gained `entries_of` — the batch sibling of `entry_of` — so the batch path asks the index what a set of cells resolves to instead of indexing its table itself.

## Why

The stated primary user does high-volume lookups against an API that answers one point per call, so every one of them writes the loop by hand — which is what `examples/batch_processing.py` used to recommend as best practice. It was the highest-ranked free item on the register: everything above it is either blocked (GH-449 on GH-542) or waiting on a maintainer decision (DATA-BINARIES on where a data release gets its binaries; GH-542's deciding figure needs a data regeneration, which a pass may not do).

## Decisions this implements

Both the maintainer's, both already on record:

- **Issue #499, 2026-08-21** — the four design questions: ids primary with names as the convenience, any array-like in with a documented zero-copy path, `on_invalid="raise"|"skip"` defaulting to `"raise"`, and `timezone_at` only in the first slice.
- **#504, in *Recorded decisions*** — a coordinate-taking interface never infers which column is which. Hence keyword `lngs`/`lats` and the refused `(N, 2)` array.

The `-1` sentinel is only safe because the negative-id guard on the public id-taking methods shipped in #551; a sentinel fed back in now raises instead of selecting a zone from the end.

## Behaviour impact

Additive for the public API — no existing signature, return type or exception changes.

The one measured effect is on the internal refactor: an ambiguous query executes exactly one more Python method call. Paired and order-alternated, 244 rounds a side, 2,000 ambiguous fixture points per round, `in_memory=False`:

| backend | min | median | head faster in |
|---|---:|---:|---|
| clang (the tracked configuration) | +0.51 % | +0.51 % | 108/244 |
| numba | +0.79 % | +1.03 % | 93/244 |

Both estimators agree on the direction, so this is real rather than noise — ~57 ns of an ~11.1 µs query on clang. Multiplied through the register's workload conversion that is **~+0.3 % of a mixed wall clock**; the unique stratum, ~89 % of a random workload, executes not one changed statement. Kept deliberately: the alternative was a second copy of the stop-index and untested-last-zone logic. Recorded in the register against the baseline anchor, so the next pass ranks on the corrected figure.

## Verification

- **Batch vs scalar over every committed fixture point**: 30,000 points x both finder classes, **0 mismatches**. Also pinned as a `slow` test in `tests/test_batch_lookup.py`, with a sampled `unit` version for the normal run.
- `make hook` clean; `make test` (1,413 passed) and `make testall` (3,222 passed) green after rebasing onto `origin/master`, which was already up to date.
- `tests/test_batch_lookup.py` also run on **Python 3.11 isolated** (56 passed) and on the **clang backend without numba**, together with `tests/test_acceleration_paths.py` (172 passed).
- `make docs` builds with no new warnings.
- `git status --short packages/timezonefinder-data/timezonefinder_data/data` empty — no data touched.
- **What the batch buys**, `clang` / `in_memory=False`, min ns per point, N = 2,000 over the committed fixtures:

  | stratum | N scalar calls | batch names | batch ids |
  |---|---:|---:|---:|
  | unique | 830 | 556 (1.49x) | 523 (**1.59x**) |
  | random | 1,791 | 1,508 (1.19x) | 1,471 (1.22x) |
  | ambiguous | 11,096 | 10,860 (1.02x) | 10,753 (1.03x) |

  With numba: 1.77x / 1.30x / 1.05x on ids. The ceiling is `h3`, which exposes no vectorised `latlng_to_cell`, so N points cost N scalar C calls whatever the rest of the pipeline does — that is why the unique stratum stops at ~1.6x and not higher. All figures are one machine's (Apple arm64, Python 3.14.2, data 2026c, fixture set v3) and are not comparable with CI's.

## Judgement calls

- **The global-function wrappers were included**, though the issue's scope line is about which lookup method gets a batch form. Every other public lookup has one, and an API surface where the batch form is reachable only through a class would be a wart.
- **No CI benchmark case.** Adding one is three lines, but `tests/test_benchmark_names.py` pins the node id set, the trend chart keys on those ids, and the report page is generated — so it obliges a `make reports` run that re-measures every committed figure on all four pages. That is a separate story; it is recorded as **BATCH-2** with the hand figures above, rather than spending half this diff on report churn.
- **`ZONE_ID_RESULT_DTYPE` is `int32`, not `int16`.** The shortcut table is `int16`, but the ambiguous fallback reads ids out of `zone_ids`, whose stored dtype is unsigned 16-bit — `int16` could not hold its upper half, and a dtype that truncates silently is worse than four bytes per answer.
- No sibling pass was in flight: the one remote `improve/*` branch is fully merged into `master`, and there were no open pull requests.

## Next

Two entries recorded in the register in place of the deleted GH-499:

- **BATCH-1** — `timezone_at_land` has no batch form, and is ranked *below* PERF-1 on purpose. Its last step is `is_ocean_timezone`, a `re.match` per answer; batching a lookup that ends in a regex per point reintroduces exactly the per-point work a batch exists to remove. After PERF-1 the check is a property of the zone id, which a batch applies as one mask over the whole answer — cheaper per point than the scalar method rather than merely equal to it.
- **BATCH-2** — nothing CI tracks measures the batch path, with the figures above and the report regeneration a benchmark case would oblige.

`certain_timezone_at` is deliberately not queued: it is a different loop that tests every candidate, and the issue's own scoping excluded it.

Refs #499


jannikmi:
## Two additions since the first review pass

### `zone_names_from_ids()` — naming a batch of ids

The batch counterpart of `zone_name_from_id`, so a pipeline can join, filter and group on ids and pay for the names once, over whatever survives.

Two regimes behind one method, because neither wins everywhere. Min ns per id, 10 % of them the sentinel:

| N | Python loop | numpy gather |
|---|---:|---:|
| 10 | 33.4 | 316.6 |
| 64 | 30.6 | 56.0 |
| **128** | **31.2** | **31.9** |
| 256 | 33.7 | 19.4 |
| 10,000 | 38.8 | 7.6 |

numpy's per-call overhead is what decides it, so the threshold is a named constant with that table beside it. The gather needs no mask: `NO_ZONE_ID` is `-1` and the lookup array carries a trailing `None`, so the sentinel indexes it by Python's own negative-index rule — the "counts from the end" behaviour the public id methods reject, used deliberately where the end *is* the sentinel. Every other negative, and the id one past the last zone (which would otherwise read that extra slot and answer `None`), are rejected first. The array is built on first use, so construction allocates nothing for it and `benchmark_results_memory.rst` still describes the tree — pinned by a test.

### Sharing a cell's preparation across the points that land in it

Points in a batch cluster, and preparing a cell — its candidate list, their zone ids, the stop index — is a property of the cell, not of the point. **2,000 points scattered around Basel reach 4 distinct candidate lists between them**, against 1,729 preparations before.

Measured before committing, because the ceiling is narrow: preparation is **898 ns against 10,228 ns** for resolving a whole ambiguous point (clang, mapped), so **8.8 %** is the most sharing can win, and only for points that actually share.

Paired and order-alternated, 240 rounds a side, min / median:

| batch workload | clang | numba |
|---|---:|---:|
| ambiguous stratum | **−3.8 % / −3.7 %** | **−6.3 % / −5.0 %** |
| clustered on Basel | −1.7 % / −1.9 % | −4.9 % / −2.8 % |
| uniformly random | −1.6 % / −0.8 % | −1.6 % / −0.3 % |
| on-land | neutral | neutral |
| unique-cell | neutral | neutral |

The clustered row is *smaller* than the ambiguous one despite ~100 % sharing, because border points carry more candidates to test, so the fixed 898 ns is a smaller share of a bigger number. Memoised by entry rather than sorted by it, so a batch that shares nothing pays a dict lookup per point instead of an `argsort`.

**The unique-cell path was checked for the same treatment and has nothing to hoist** — it is already a single `np.where` over the whole batch with no per-point Python at all, and `unique_shortcut_points` contains zero ambiguous points by construction. The measurement agrees: that stratum moves −0.3 % / −0.8 %, i.e. not at all.

**The single-point path is left alone.** `_zone_id_in_ambiguous_cell` writes its three preparation expressions out rather than calling the shared helper: **−0.41 % min / −0.08 % median over 244 rounds** against the commit before it. That three-expression duplication cannot drift silently — `test_batch_and_scalar_agree_over_every_committed_point` compares both paths over every point of every committed fixture.

### Correction to a figure in the PR description

The `+0.51 %` quoted above for the shared candidate loop's cost to the scalar path does not reproduce to that precision. A three-way interleaved re-run (master / pre-grouping / grouped, 244 rounds each in one session) put the same comparison at +2.12 % min / +0.70 % median, estimators straddling. **The durable fact is the count — exactly one added method call — and the time is order 1 % of an ambiguous query, ~0.6 % of a mixed wall clock.** `potential-improvements.md` now says that rather than a second decimal this machine cannot support.

### Verification

`make hook` clean; `make testall` **3,271 passed**; batch-vs-scalar agreement re-checked over all 30,000 committed fixture points × both finder classes with **0 mismatches**; rebased onto `origin/master`, which was already up to date.

PR #555, opened 2026-08-24
Measure how far from a border tzfpy and timezonefinder stop agreeing

jannikmi:
## What

The competitor half of **GH-542** (issue #542). The issue was opened to unblock the polygon-encoding choice in #449; #562 has since replaced that encoding with FMT-1/FMT-2, so this PR now preserves the precision evidence without claiming to gate the replacement. `make tzfpy-agreement` measures the effect of `tzfpy`'s simplified boundaries at stated, globally verified distances from the nearest timezone border.

The primary metric samples a border location uniformly by edge length, verifies a probe on each side, and counts the location once when either answer differs. This gives the impact an honest 0–100% scale even when a displaced boundary changes answers on only one side. The original individual-point probability over the complete two-sided offset locus remains in the report as a secondary metric.

affected border locations

Boundary release 2026c on both sides, `tzfpy` 1.3.3, 20,000 accepted paired locations and 20,000 accepted individual points per distance:

| Distance | Any border location | Land-zone border location | One side | Both sides | Not attributed |
|---|---:|---:|---:|---:|---:|
| 1 cm | 71.8% | 89.6% | 69.0% | 2.79% | 19 |
| 5 cm | 69.1% | 90.5% | 69.0% | 0.045% | 0 |
| 10 cm | 66.9% | 87.7% | 66.9% | 0.035% | 0 |
| 50 cm | 54.4% | 71.4% | 54.4% | 0.025% | 0 |
| 1 m | 50.8% | 67.1% | 50.8% | 0.015% | 0 |
| 5 m | 43.0% | 56.2% | 43.0% | 0.010% | 0 |
| 10 m | 36.0% | 47.5% | 36.0% | 0 | 0 |
| 50 m | 8.63% | 11.6% | 8.63% | 0 | 0 |
| 100 m | 0.20% | 0.27% | 0.20% | 0 | 0 |
| 500 m | 0 observed of 20,000 | 0 observed | 0 | 0 | 0 |
| 1 km | 0 observed of 20,000 | 0 observed | 0 | 0 | 0 |

The 1 cm row lies just below timezonefinder's roughly 1.1 cm stored-coordinate step, so 5 cm is the clean near-border headline: 90.5% of sampled land-zone border locations are affected. The SVG labels decade distances only; the 500 m upper bound is deliberately not duplicated beside 1 km. Display precision follows what the sample supports: tenths for ordinary rates, hundredths below 1%, and thousandths below 0.1%.

## Attribution and BUG-3

[...]

Issue #522, opened 2026-08-18
Shrink the repository history by dropping the committed coordinate binaries


jannikmi:
Splitting off the data distribution (#446) stops the repository *growing*. It does not shrink
what is already there — deleting a file leaves every past blob in the history, so a clone keeps
transferring the same bytes. Reclaiming them needs a history rewrite, which is a separate
decision with its own blast radius. This issue tracks it.

## What it would reclaim

Deduplicating blobs across all history:

| | |
|---|---|
| all unique blobs | 571.4 MiB |
| coordinate binaries | 543.8 MiB (**95%**) |

That is `boundaries/coordinates.fbs` and its predecessors `poly_data.bin` and
`timezone_data.bin` — 46 distinct blobs. The pack is currently 356.98 MiB; dropping them takes
the repository to single-digit MiB.

For reference, in the current tree `coordinates.fbs` is 63.4 MB of the 65.2 MB under
`timezonefinder/data`. Everything else totals 1.79 MB and is not worth rewriting history over.

## Sequencing: this comes after #446

The rewrite has to land *after* the "no binaries committed in future" decision from #446, not
before. Rewriting first and then taking another data update through the current pipeline re-adds
~62 MB immediately, and the rewrite would have to be repeated.

## What it costs

`git filter-repo` rewrites every commit SHA. Consequences:

- every existing clone and fork is detached from the new history and has to be re-cloned
- all tags are rewritten; any signatures on them are invalidated
- links to commit SHAs from issues, PRs, changelogs and external references break
- it is a force-push over `master` and every tag

None of that is recoverable by halves, so it is worth doing exactly once — which is the other
reason to wait for #446 to settle first.

## Open questions

- Whether the reclaimed size justifies the disruption at all, given that `actions/checkout` is
  shallow by default and CI therefore never paid for the history. The cost falls on humans
  cloning the repository and on anyone with an existing fork.
- Whether to preserve the old history somewhere immutable (an archived repository, or a tag on a
  detached pre-rewrite ref) so historical data versions remain retrievable.

Issue #513, opened 2026-08-14
Drop hole polygons entirely, once shortcut candidate ordering can guarantee it


jannikmi:
Follow-up to #350 / #509. **The measurement below is the point of this issue: dropping holes today
returns wrong timezones, so this is blocked on establishing an ordering guarantee, not on verifying
one that already holds.**

## The idea

Every hole exists to say "this area is *not* the surrounding zone". But the area is always some
other zone's polygon, and the lookup already tests those polygons. If a covering polygon is
guaranteed to be tested before the surrounding one, the hole carries no information and can be
dropped — along with the entire parallel data structure that serves it.

That is the end state the original #350 description assumed. #509 delivered only the storage half
(holes that duplicate a boundary polygon are stored as a reference to it); the structural half is
this issue.

## What it would remove

| | |
|---|---|
| `timezonefinder/data/holes/` (6 files) | ~180 KB |
| `timezonefinder/data/hole_registry.json` | ~3 KB |
| `HoleArray` (`polygon_array.py`) | ~85 lines |
| `_iter_hole_ids_of`, `_holes_of_poly`, `_load_hole_registry`, `hole_registry`, `nr_of_holes` | `timezonefinder.py` |
| the holes-before-boundary branch in `inside_of_polygon` | one hole test per candidate polygon, on the query path |
| `poly_ref.npy` and the whole reference encoding from #509 | superseded |
| `HoleCollection`, its deduplication pass and `scripts/data_integrity.py`'s hole checks | build side |

Note the last two rows: **if this lands, #509's encoding is deleted, not extended.** #509 is still
worth having in the meantime — it is a pure storage win available today, whereas this is not — but
nothing should be built on top of the reference vector on the assumption it is permanent.

## Why it is blocked — measured, not assumed

I built two variants of the packaged data by rewriting only the hole files, and compared
`timezone_at` against the packaged data over 6,048 interior points of holes (8 per hole, rejection
sampled) plus 20,000 uniformly random global points.

| variant | hole interior points changed | holes affected | random global points changed |
|---|---|---|---|
| drop only the 27 holes with no boundary twin | 160 / 6,048 | 20 of 27 | 0 / 20,000 |
| drop **all** holes | 1,703 / 6,048 | 224 of 756 | 16 / 20,000 |

The changed answers are wrong, not merely different:

```
drop the 27:            152  Asia/Hebron               ->  Asia/Jerusalem
[...]

PR #563, opened 2026-08-30
A latitude block index over the boundary polygons

jannikmi:
## What this is

FMT-1, the top-ranked free item: a **latitude block index** over the boundary polygons. Every stored ring is split into blocks of 128 vertices, and the range of latitudes each block's edges span is stored beside the coordinates. Ray casting flips parity only on an edge that spans the query latitude, so a block whose range excludes it holds nothing that can change the answer and is skipped whole.

The coordinate payload is unchanged byte for byte apart from the ring rotation below — this adds an index *beside* the coordinates rather than re-encoding them, which is what makes it separable from FMT-2.

## Why this item

It was the highest-ranked free row and the only thing gating the held format-2 window, and with it the 2.x data release, DATA-BINARIES, GH-522 and the whole pending code release. Nothing above it was eligible: DATA-BINARIES was blocked behind exactly this, and FMT-2 was blocked behind its block structure.

## Behaviour impact

- **Answers are unchanged.** The skip is exact, not approximate, and that is checked rather than argued — see *Verification*.
- **`get_geometry()` returns each ring starting at a different vertex** than the upstream GeoJSON did. The converter now chooses each stored ring's start index to make its blocks tightest; it is the same closed path, and nothing downstream depends on where a ring begins (`canonical_ring_key` is rotation-invariant, bounding boxes are unaffected, a referenced hole follows its boundary). Worth ~3.9 % of the edges a query scans — taken as a free step of a rebuild that was happening anyway, never claimed as a speedup.
- **`POLYGON_LAYOUT_VERSION` → 2**, and `DATA_FORMAT_LAYOUT_VERSIONS` with it. **`DATA_FORMAT_VERSION` stays at 2** — that is what the held window was for. This needs the `timezonefinder-data` 2.x release the pending bump already required; it does not add a release.
- `utils.inside_polygon_blocked` joins `utils.inside_polygon` as a second import-time binding. The bare kernel stays, because build-time geometry code and the kernel tests legitimately hold a ring with no stored index. `using_clang_pip()` now reports the blocked binding, which is the one a lookup reaches.
- **Memory:** +~0.5 MB on the heap in either mode (the index is read in, like the bbox vectors). The mapped mode's *resident* set after a workload falls ~26.6 → ~17.2 MiB, because a lookup that skips a block never faults in the coordinate pages behind it.

## Measurement

Paired, order-alternated, in **one process** on the **C extension** in the **default mapped mode** — only `PolygonArray.pip` swapped, same data, same interpreter. 41 rounds × 2,000 points, reproduced three times:

| stratum | µs/query, min | rounds won | p99 | p99.9 |
|---|---|---|---|---|
| random | 1.92 → 1.51 (**−20 %**) | 41 / 41 | 26.0 → 12.1 | 46.3 → 20.3 |
| on_land | 3.66 → 2.09 (**−42 %**) | 41 / 41 | 36.2 → 12.5 | 81.5 → 30.1 |
| ambiguous | 11.69 → 6.55 (**−42 %**) | 41 / 41 | 55.2 → 18.8 | 86.1 → 35.5 |
| unique | 0.95 → 0.95 (+0.3 %) | 27 / 41 → no effect | 1.33 → 1.29 | 3.21 → 2.67 |

Both estimators agree on the three strata that read geometry, and the median does not move. The unique stratum never reaches a point-in-polygon test, so it is expected to be untouched — and is.

The mechanism, from `prototypes/query_stage_profile.py`: **the kernel goes from linear in polygon size to almost flat.** Over a 46,823-vertex ring it costs **591 ns against 21,462 ns**; over a 112-vertex ring it is unchanged. An ambiguous query is 11,688 → 6,791 ns, `on_land` 3,882 → 2,095 ns, `unique` 970 → 976 ns.

**The one thing this makes worse** is a small polygon, by ~400 ns per test: the index slice and, on clang, a third `ffi.from_buffer` buy nothing for a ring that fits in one block, since the bbox check has already rejected everything outside it. 29.7 % of polygons are in that class. It is ~3 % of an ambiguous query — below what any benchmark here resolves — and the whole-query A/B is decisive anyway, so it is **recorded as PERF-7 rather than fixed with a branch**.

## Verification

- `make testall` — **3,389 passed**, on the committed tree after rebasing onto `origin/master` (which had not moved).
- `uv run pre-commit run --all-files` — all hooks pass. (`make hook` cannot run here: it refuses to install with `core.hooksPath` set.)
- The clang axis run explicitly: `uv run --isolated --group test pytest tests/test_block_index.py tests/utils_test.py tests/test_hole_references.py` — 121 passed.
- `tests/test_block_index.py` is the new seam. It replays **every point-in-polygon test the committed query fixtures actually produce** — recorded off a real lookup, not constructed — and asserts the blocked kernel answers exactly what the unblocked one does, on both backends. It also asserts the filter *does* something (it leaves 1.8 % of the edges), that every hole is indexed by the ring it resolves to, that every edge lies inside its own block's range, and that the integrity check rejects a narrowed range and a different block size.
- `scripts.data_integrity.validate_block_index` re-derives every ring's ranges from the stored coordinates — run by the converter over what it wrote and by the test suite over what ships.
- Data diff lists **only** the two coordinate files and the four new index files; `shortcuts.bin`, the bbox vectors, `zone_ids.npy` and `poly_ref.npy` are byte-identical.
- All four benchmark report pages, the memory page and the profiler's `FINDINGS` are re-measured against this tree (Darwin arm64, Python 3.14.2, C extension, mapped — not comparable with CI's).

## Judgment calls

- **The index lives in two `.npy` files beside `coordinates.bin`, not inside it.** `poly_ref.npy` is the precedent: a per-polygon vector covered by the polygon layout version. It is read into memory in both modes, deliberately — the point of the index is to decide *without* touching the coordinates which pages have to be faulted in, so mapping it would spend the fetch it exists to avoid.
- **`POLYGON_BLOCK_SIZE` is a code constant, not a header field.** Nothing in the files records the size they were built at, so it is part of what the layout version means; a directory blocked at another size is rejected by the layout marker and, loudly, by the integrity check. `scripts/tune_block_size.py` re-runs the sweep: 128 stands, 64 ties it on work units at twice the index size, 32/256/512 all cost more.
- **Two kernels per backend rather than one shared body.** The two are already two copies of the same predicate (numba and C), kept honest by tests rather than by sharing; adding blocked variants follows that. The alternative — one kernel with a degenerate block — costs an allocation per call on the bare path.
- **`--latency-json` is required by `scripts.render_benchmark_reports`**, unlike `--memory-json`, because the distribution is a *section* of a page the renderer rewrites wholesale. Both contributor-memory statements of that command are updated.
- **`.gitignore` gains `timezonefinder/inside_polygon_ext.c`** — a cffi build artifact that appears whenever the extension is compiled in place, which this change makes people do. One line, and I hit exactly this trap while writing it.

## Concurrent work

PR #555 (`improve/542-tzfpy-agreement`) is open and touches the GH-542 entry and the ranking table. I fetched before starting and again before the final gate; `origin/master` had not moved. **Whichever of the two merges second takes the other's rows** — the collision is one line each in `improvement-priority-ranking.md`.

## Next eligible item

**FMT-2** — the frame-of-reference payload. Its blocker was this change's block structure, which now exists; it takes a `DATA_FORMAT_VERSION` 3, which GH-513 should share. Its open question (bit-packed against byte-aligned widths) can now be measured inside the shipped blocked kernel.

DATA-BINARIES remains blocked, but only on the 2.x publication itself rather than on another change.


jannikmi:
## CI's second machine class agrees

The benchmark job above is the cross-check the [measurement baseline](https://github.com/jannikmi/timezonefinder/blob/master/contributing/improvements/query-performance-measurement-baseline.md) asks for: base and head in **one job on one runner**, Linux on an AMD EPYC 9V74, C extension without numba — the tracked configuration, and a different machine class from the Darwin arm64 figures in the description.

| stratum | this laptop (paired, in-process) | CI (same-runner base vs head) |
|---|---|---|
| random | −20 % | **−18.3 %** |
| ambiguous | −42 % | **−41.3 %** |
| unique | no effect (27 / 41 rounds) | +8.2 %, flagged unchanged |

The two agree to within a couple of points on the strata that read geometry, so the result is not a property of one machine.

**On the `unique_shortcut` row**, since it is the one that does not agree cleanly: it sits under the 110 % flag threshold, and there is no code-path change behind it — a unique-cell query is answered by a table read and never reaches a point-in-polygon test, so the block index cannot be executed on it. An in-process paired A/B on this laptop puts it at +0.3 % with the two estimators disagreeing, and an alternating base/head comparison pinned to the same interpreter reads 986–990 ns against 985–1001 ns. What is left that could be real is layout rather than instructions: the finder holds ~0.5 MB more, which can move cache behaviour for the shortcut table. Worth re-reading on the next benchmark run rather than asserting it is zero.

**A measurement trap this turned up**, recorded in the profiler's `FINDINGS` because it cost real time: `uv` resolves `.python-version` by walking up from the working directory, so a comparison worktree created *outside* the repository silently gets a different CPython — here a free-threaded build against the pinned Homebrew one. Pure-Python call chains differ ~50 % between them, which presented as a 17 % regression on exactly this stratum until both sides were pinned with `--python`.

jannikmi:
Thanks — the rotation finding is correct, and it was an assumption I inherited from the prototype without checking. Verified, measured, and answered in a26fda1.

**The mechanism is exactly as described.** Rotating by a whole block only relabels the blocks when the block size divides the vertex count; otherwise it moves the ragged final block and repartitions the ring. So there are `n` distinct rotations, not `block_size`. Reproduced on the packaged data: for rings of 129–1,000 vertices, the bounded search misses the true minimum for **193 of 544**, by up to **15.97 %** of the span sum — and never for the 5 rings whose vertex count `B` divides, which is the mechanism confirming itself.

**Exhaustive search is affordable.** Every rotation's span sum is available in O(n) from sliding-window extrema plus the recurrence `S(r+B) = S(r) − span_B[r] + span_B[p] − span_L[p] + span_L[r+d]` where `p = r+(nb−1)B` and `d = nb·B−n`. Verified against brute force over 80 random rings: 0 mismatches. **4.4 s for the whole collection** against ~2 s for the bounded sweep, in a converter that already takes ~56 s.

**And it is refused anyway, because it loses on the goal.** Over the real (point, polygon) pairs the committed fixtures produce:

| rotation | edges scanned | |
|---|---:|---|
| as shipped (bounded search) | 1,332,544 | 1.0000x |
| exact, all `n` rotations | 1,334,335 | **1.0013x** |

The exact search improves the *span sum* for 392 of the 602 rings reached and makes the shipped data marginally **worse** to query. The span sum is a proxy — it is the expected edges scanned for a latitude drawn *uniformly from the ring's own range*, and real query latitudes are not distributed that way. Past a point, optimising the proxy stops tracking the goal. Both figures sit far inside the 3–9 % noise floor, so the honest reading is that the two are indistinguishable in effect and the bounded search is less code.

**What changed as a result**, since the defect in the claim was real even though the fix is not worth making:

- `best_rotation_offset` now states that the bound is a heuristic, with the counts above and a pointer to the refusal. The `range(block_size)` is commented as deliberate.
- `docs/data_format.rst` and `prototypes/polygon_block_encoding.py` said "only offsets in [0, B) differ" in three places. All corrected; the prototype gains finding 9 with the recurrence and the numbers.
- The refusal is recorded in the geometry decisions memory with its evidence, so this is not re-raised as a defect on a future pass. The generalisable half is stated there too: anything proposing to minimise this objective more exactly has to show a gain in *edges scanned over real query pairs*, not in the span sum.

`make testall` (3,389 passed) and `pre-commit run --all-files` are green.

jannikmi:
## Following the rotation finding further changed the answer

My first response kept the bounded search and recorded a refusal. That was wrong, and chasing the reviewer's point properly turned up *why* it was hard to act on: the search space was too small **and the objective was wrong**, and both have the same cause — the final block is ragged.

Expected edges scanned is `Σ size_b · span_b / total_span`. The old objective summed spans **unweighted**, so a ragged block holding one edge counted exactly as much as a full block holding 128, and the search would happily inflate a real cost to shrink a negligible one. Rings whose vertex count the block size does not divide are the common case, not an edge case — which is exactly the population where the bounded search was also missing the minimum.

Measured over edges actually scanned on the real query pairs, against what shipped:

| change | edges scanned | |
|---|---:|---|
| shipped (unweighted spans, bounded to one block) | 1,332,544 | 1.0000x |
| **fix the search only** | 1,334,335 | **1.0013x** |
| fix the weighting only | 1,311,617 | 0.9843x |
| **fix both** | 1,253,338 | **0.9406x** |

That second row is the instructive one and it is why my first answer read the way it did: widening a search against a wrong objective finds a better answer to the wrong question. Once the objective is right, the wider search is worth 4.4 percentage points more.

**Both are now implemented.** Searching every rotation would be quadratic done directly; it is linear in the block count instead — every window's span is computed once for all start positions, and each rotation is a gather and a sum over those, chunked so the index array stays small. 12 s for the whole collection against 2 s, in a converter that takes about a minute. Verified optimal against brute force over 140 random rings.

The data is regenerated and everything measured on it re-run. The whole-query A/B is unchanged within noise (random −20.7 %, on_land −43.6 %, ambiguous −42.7 %, unique neutral), all four report pages, the memory page and the profiler `FINDINGS` are re-measured, and the block-size sweep still puts 128 at the optimum (64 ties on work units at twice the index size).

The decision record now carries the corrected story rather than the refusal, including the generalisable half: **before widening a search, check that the objective is the thing you want minimised** — and state an expected cost as a weighted sum over the units that actually cost, because a count of blocks is not a count of edges.

Rebased onto `master` (#564 landed); `make testall` 3,398 passed, hooks clean.

jannikmi:
@ringsaturn have a look. this about the "long-tail latency" you mentioned.
This improvement will ship in the next release.

PR #565, opened 2026-08-30
Report a second estimator beside the tracked one in the benchmark comparison

jannikmi:
## What changed

The pull request benchmark comparison reduced each side to a single `min` and threw away every other statistic pytest-benchmark had already recorded. That left the table unable to express the one answer a reviewer most often needs — *no demonstrable difference* — because a single estimator can only ever report a number and a verdict derived from it.

Every row of `scripts/compare_benchmark_runs.py`'s table now carries its `median` change beside the tracked `min` change, and is marked `⚠️ unresolved` where the two land on different sides of the flag. `median` is the corroborating statistic because `scripts/normalize_benchmark_json.py` overwrites `mean` with the tracked value, leaving `median` the only untouched estimator in a stored report.

This is BENCH-1's first listed fix, the one its entry calls "most of the value". The item is rewritten rather than deleted: the paired A/B harness for comparing two candidate implementations of one stage is still open, and is what the entry now describes.

## Why this item

BENCH-1 is the highest-ranked eligible row. FMT-1 is claimed by #563; DATA-BINARIES and FMT-2 are blocked by it; GH-542 is claimed by #555. Nothing was batched with it: every remaining eligible row is medium or broader, or touches runtime files #563 is already changing.

## Behavior impact

The flag is unchanged and still derived from the tracked estimator, so what CI reports for a real change is the same and `--fail-on-regression` remains unused, exactly as the item requires ("whatever is added stays reporting-only"). Two smaller judgment calls:

- The "faster" side of the band moves from `factor >= threshold` (−9.09 %) to a ratio reaching the threshold in either direction (−10 %), so the two directions are now called out on identical evidence. Both sides of that band sit inside the noise floor, so no row this changes was ever actionable.
- A report carrying only one statistic — an older one, or a metric that records one — drops the column rather than failing the comparison the workflow depends on. The memory comparison, which shares this machinery, does record `median` and gets the column.

Nothing in the package changes. Packaged data diff is empty.

## Verification

- untouched baselines on `e9afb89`: `make test` — 1,535 passed; `pre-commit run --all-files` — passed
- **real end-to-end run, not just unit tests**: `make benchmarks-ci` twice on identical code, then compared. Both sides on one Apple M1 Pro; the tracked `min` moved +0.1 / −0.4 / −0.5 % while the corroborating `median` moved +3.1 / +2.1 / +0.2 % — all three rows `⚪ unchanged`, which is the correct answer for unchanged code, and a worked demonstration that the two estimators do not track each other inside the noise band
- memory metric path rendered against a synthetic `scripts.measure_memory` report — column present and populated
- targeted: `uv run pytest tests/test_benchmark_ci_tooling.py tests/test_benchmark_workflows.py tests/test_improvement_ledger.py` — 54 passed
- final gate, after rebasing onto `323ae2f` (#564 merged mid-pass): `pre-commit run --all-files` — passed; `make test` — 1,544 passed; `make testall` — 3,363 passed; `make docs` — build succeeded

## Decisions

None needed. Reporting a statistic the report already carries, beside one already reported, is reversible and does not change what CI passes or fails on.

## Concurrent work

Yielded to #563 (FMT-1) and #555 (GH-542). #564 merged during this pass; the branch was rebased onto it and the whole gate re-run, resolving one `CHANGELOG.rst` conflict by keeping both sides.

## Next

BENCH-1 remains the top free row, now describing only the paired A/B harness. BATCH-2 follows it.


jannikmi:
Taken, and it was right — the register's own rule is "It is a to-do list, not a history", so a `Shipped` bullet is exactly what should not be there.

Rather than only deleting that bullet, the entry is rewritten around its one remaining subject: a reusable, order-alternating harness for comparing two *candidate implementations* of one stage. That is a different subject from the base/head comparison the entry was named after, so the heading, the filename and the ranking row now say so too. What survives is the evidence that argues *for* the harness — the −13.3 % → −0.3 % warm-up result, the isolated-stage and dict-ordering designs, PERF-4 — none of which is shipped history; it is the case for the open work, and losing it would cost the next pass a rediscovery.

Two consequences the rename obliged, both in this commit:

- FMT-1 described BENCH-1 as "about base-versus-head *comparison* tooling drifting between two runs", which is no longer true. Rewritten to the lasting fact, including that the per-query harness FMT-1 owes is not this harness.
- The old filename is gone, so the single inbound link (the ranking row) moved with it. `tests/test_improvement_ledger.py` passes, which is what enforces the row ↔ entry pairing.

Unrelated to your comment but worth stating for anyone reading this PR's own benchmark comment: it shows the **old** five-column table. That is not a defect. `benchmark-comment.yml` is a `workflow_run` job, so it runs the default branch's copy of `scripts/compare_benchmark_runs.py` and never the PR's — deliberately, as its header explains, because the only thing crossing the trust boundary is the benchmark JSON. The new column appears on the first PR opened after this merges. That non-obvious fact is now recorded in `contributing/development/benchmarking-and-performance-validation.md`, since reading the old table as evidence the change is broken is the obvious failure mode.

Re-verified after the rewrite: `make hook` passed, `make test` 1,544 passed, `make testall` 3,363 passed, `make docs` built clean, packaged data diff empty.

One overlap worth flagging for merge order: #566 changes `compare_runs` and `render_markdown` in the same file, to tolerate benchmarks present on only one side (BATCH-2 adds cases the merge base lacks). The two changes are complementary rather than competing — this one adds a column, that one relaxes the set check — but they touch the same two functions, so whichever lands second needs a rebase rather than a merge resolution taken on autopilot.

jannikmi:
@codex review

Issue #500, opened 2026-08-13
Validate a data directory's cross-file invariants (`--validate-data`)


jannikmi:
`TimezoneFinder.timezone_at` takes two shortcuts that are correct only because of invariants the
reader never checks:

```python
for i, boundary_id in enumerate(possible_boundaries):
    if i >= last_zone_change_idx:
        break                      # (1) stop early
    ...
zone_id = zone_ids[-1]
return self.zone_name_from_id(int(zone_id))   # (2) return the last candidate UNTESTED
```

Both are sound **only if** the shortcut's candidate list is complete and grouped by zone id. That
grouping is asserted by `scripts/shortcuts.py::check_shortcut_sorting` — at **generation** time.
Nothing re-checks it when data is loaded.

`bin_file_location` is documented public API and `docs/2_use_cases.rst` actively points users at
compiling their own data. So a user whose pipeline differs subtly gets **silently wrong timezones**,
with no error and no way to notice.

This is the same failure mode that the coordinate-layout guard just closed — per that changelog
entry, such a directory was previously *"read without complaint and produced wrong timezones"*.
#458 is the next narrow slice (the shortcut file format marker). This issue is the general case:
every invariant the fast path assumes is currently enforced by the generator and by nothing the
reader can see.

## Proposal

A `validate_data_dir(path)` function plus a `timezonefinder --validate-data <dir>` CLI mode,
checking at minimum:

- `zone_positions` monotonic non-decreasing and correctly terminated
- `zone_ids` within `[0, nr_of_zones)` and non-decreasing
- **shortcut candidate lists grouped by zone id** (the invariant behind (1) and (2) above)
- every shortcut polygon id within range
- per-polygon bounding boxes actually containing their polygons
- `hole_registry` references resolvable, hole ids in range
- file identifiers and layout versions present and matching (overlaps #458)

## Design notes

- **Opt-in, not on construction.** Initialisation time is a tracked benchmark
  (`benchmark_results_initialization`), so validating eagerly on every `TimezoneFinder()` is the
  wrong default.
- **Also run it at the end of `scripts/file_converter.py`**, so the generator self-checks its own
  output — that is free and closes the loop.
- **Share the CLI design with #428**, which proposes `timezonefinder update_data`. Two subcommands
  landing separately will produce two different conventions.

Purely additive. The main risk is discovering that the shipped data violates an invariant the code
has been quietly tolerating — which is a feature of the exercise, not a reason to defer it.

[...]

PR #576, opened 2026-08-31
Measure the impact of reducing coordinate precision

### LINKED ISSUE #542: Establish what coordinate precision is worth: what competitors offer, and what users need

jannikmi:
## Selection

At the start of this pass, DATA-BINARIES was blocked on publication of the 2.x data package and FMT-2 had an authoritative improvement claim, making GH-542 the highest-ranked eligible item. BENCH-1 and BATCH-2 were not batched because they exercise separate CI/reporting machinery, and BATCH-2 still needs runner-noise calibration.

DATA-BINARIES PRs #571 and #572 merged while this pass ran. This branch was rebased onto both changes and preserves their bootstrap and one-wheel hand-off records. DATA-BINARIES and FMT-2 now have active PRs #575 and #574; GH-500 remains active in #573. After those in-flight higher-ranked items, BENCH-1 is the next eligible improvement.

## What changed

- add a deterministic measurement tool that rebuilds the complete packaged geometry and H3 shortcut index at five decimal places, then compares it with the six-decimal source geometry
- sample paired, verified border probes at 10 cm, 50 cm, 1 m, and 5 m, plus an area-uniform globe population and every committed benchmark fixture class
- commit the machine-readable run and generated documentation, and make the report reproducible through `make precision-impact` / `make reports`
- record the durable conclusion that six source decimal places are the accuracy floor; only the package representation's redundant seventh decimal may be omitted
- remove the completed GH-542 register item in a separate commit

The runtime data and generated FlatBuffers bindings are unchanged.

## Result

Five-decimal nearest rounding changes real lookup answers near borders:

| Distance from border | Any-border locations affected | Land-border locations affected |
|---|---:|---:|
| 10 cm | 4,876 / 10,000 (48.8%) | 4,873 / 7,597 (64.1%) |
| 50 cm | 231 / 10,000 (2.3%) | 231 / 7,548 (3.1%) |
| 1 m | 0 / 10,000 (95% bound <0.030%) | 0 / 7,735 (95% bound <0.039%) |
| 5 m | 0 / 10,000 (95% bound <0.030%) | 0 / 7,659 (95% bound <0.039%) |

The area-uniform globe sample changed 0 / 200,000 answers (95% bound <0.002%), and all 30,000 committed benchmark fixture points were unchanged. Those ordinary-workload results do not erase the targeted border result: the source's sixth decimal is load-bearing.

No maintainer decision was required. The contributor-level experimental choice was nearest rounding, ties away from zero, because it gives the coarser candidate its smallest displacement. Rebuilding the shortcut index was required so both candidate lists and polygons describe the same geometry.

## Verification

- `make hook`
- `make test` — 1,730 passed, 1,848 deselected
- `make docs` — Sphinx build succeeded, including the JSON download
- `make testall` — 3,578 passed
- `git diff --check origin/master...HEAD`

Closes #542


jannikmi:
sticking with the source precision, as "no border simplification" is the main claim of this package.
The tooling of this PR is not required.

PR #578, opened 2026-08-31
Compile a branch's packaged data in CI instead of committing it

jannikmi:
Follow-up to #575, which git-ignored the packaged dataset and left one gap behind: a branch whose data version PyPI has never served had nothing to hand CI, so `git add -f` and ~63 MB of history per format change was the documented answer. This closes it.

## What changed

**`.github/workflows/compile_data.yml`** — dispatched on a branch, compiles that branch's dataset and uploads it as `artifact-data-wheel`:

```bash
gh workflow run compile_data.yml --ref <branch>
```

The run reports its id in the job summary; commit it as `DATA_BUILD_RUN` and every consumer already knows what to do — the tox matrix, both benchmark checkouts, `make-data-wheel` and the `data-v*` tag all read the recorded run before they reach for PyPI. That machinery shipped in #575; only the producer was missing.

**`update_data.sh` grew the two flags it needs**, both defaulting off:

- `--tag=<release>` pins the release instead of resolving upstream. The verification against the release API is unchanged, so a tag that does not exist fails there rather than compiling something unattributed — checked against a bogus tag, which never reaches the converter.
- `--binaries-only` stops after the converter: no stamps, no version bump, no fixtures, no reports.

## Why it pins the release rather than resolving latest

This is the part worth reviewing. The branch's committed benchmark fixtures and its five report pages are pinned to `DATA_VERSION`. Compiling whatever is latest would not produce a *broken* dataset — it would produce one that installs, answers lookups, and silently disagrees with every figure the branch published. So the tag is read out of the tree, and `test_the_dispatch_compiles_the_release_its_branch_declares` pins that.

`--binaries-only` is the same argument from the other side: the full run would overwrite the branch's measured report pages with a runner's numbers.

## Why not just commit the binaries

Recorded as a refused option in the [format-release decisions](contributing/improvements/decisions/data-format-version-marker-and-release-batching-decisions.md), because it needs no machinery at all and will otherwise be re-proposed on exactly that basis. It costs ~63 MB of history per format change — reintroducing the growth #575 removed — and it returns the release to vouching for laptop-compiled bytes with a diff. It still physically works, since ignoring never affects a tracked file; it is the fallback, not the plan.

The stronger argument is the one it loses on: the wheel this run builds is the wheel the pull request's matrix installs and the wheel the tag publishes. A format change now reaches PyPI through **one build**, which is the invariant `publish_data.yml` states in its own header and which the weekly pipeline already had. Committing binaries is the only path that breaks it.

## Behaviour impact

None at runtime, and none for the weekly pipeline — both new flags default off, and `test_the_update_script_still_prepares_a_release_by_default` pins each step of the tail they skip. The published wheels are unchanged.

GH-522's caveat is gone: its entry no longer has to say "run the rewrite when no such branch is unmerged", because nothing adds a 47th coordinate blob.

## Verification

- `uv run pre-commit run --all-files` clean; `make testall` **3,579 passed**
- `bash update_data.sh --tag=9999z --with-oceans --binaries-only` → `PINNED TO RELEASE 9999z`, the tag lands in the download URL, 404, exit 8. The pin flows through and no "resolve latest" call is made
- `bash -n` over every `run:` block in `.github/`
- the widened builder scan proven non-vacuous: it reaches all 7 workflow files and reports both producers

**The dispatch itself has not been run.** It cannot be until this is on a branch GitHub will dispatch, and the conversion has never executed in CI at all — all 30 recent `check_data_updates.yml` runs skipped the regeneration job, since upstream has published nothing since that pipeline was built. The 180-minute timeout is inherited from that job and is an unmeasured ceiling, not a figure. The first dispatch — most usefully on #574 — is what measures it.

## Tests

The handoff tests were scanning a hardcoded triple of workflows, so a second builder in a *new file* was invisible to precisely the check meant to catch one. They now enumerate the workflow directory and pin both producers by name, plus:

- both producers upload under one artefact name, and both actions download that name
- the dispatch pins `DATA_VERSION` and passes `--binaries-only`
- `update_data.sh` still defaults to resolving latest and still runs every step of the release preparation

## Judgment calls

- **Extended `update_data.sh` rather than writing a second compiler.** It is the script an auto-merging pipeline runs unattended, so touching it is not free — but the alternative duplicates the digest verification and the release-asset naming, which is exactly the drift this repository keeps testing for. Both flags subtract, both default off, and the default path is pinned step by step.
- **The decision went in the format-release file, not the distribution one.** The natural home was at 1,996 of its 2,000-word budget, and trimming another recorded decision to fit this one would have degraded memory to satisfy a limit. The format-release file is the right topic anyway — it already carries what a format change costs and why FMT-2 takes a format 3.
- **90-day artefact retention, set explicitly.** The weekly update's wheel is consumed within minutes by a pull request that auto-merges; this one has to survive the review of a format change. If it expires, re-dispatch — the run id is the only reference to it.
- **A lookup assertion before the upload.** The same one the end-to-end job makes of a released wheel. On a format branch it is the branch's own reader pointed at the branch's own data for the first time, which is worth finding here rather than four jobs later.

## Concurrent work

#577 restructures `improvement-discovery-coverage-map.md`, which this touches in one sentence — a conflict to expect, trivially resolved. #576 and #573 touch the register and the decisions tree but not the files edited here. #574 is the intended first user.

## Next

The next eligible item in the ranking is **FMT-2** (#574). With this merged it needs no `git add -f`: dispatch, commit the run id, and its binaries never enter the repository.


jannikmi:
Good catch — the flag's promise was not true. Fixed in d0f38de.

**Confirmed, and bounded.** `parse_data` calls `write_data_report_from_binary` unconditionally, and that writes the checkout's `docs/data_report.rst` whatever `-out` it was given. Measured rather than assumed: a converter run over the test fixture on a clean tree dirties that one tracked file and nothing else. So a `--binaries-only` run left a modified tracked file behind while saying it regenerates no reports, and on a runner it left the dispatch's checkout dirty for whatever step looked next.

**The fix restores that file rather than suppressing the converter.** `--binaries-only` puts `docs/data_report.rst` back after the parse, in the checkout it ran from, guarded on actually being in a git repository.

Not the converter change you suggested, deliberately: making `parse_data` respect its `-out` is [TOOL-6](https://github.com/jannikmi/timezonefinder/blob/master/contributing/improvements/items/data-pipeline-and-developer-tooling/tool-6-parse-data-rewrites-the-committed-data-report-whatever-out-it-was-given.md), a separately ranked and already-decided item with its own design and its own ~150-line diff. Doing it from here would pull an unrelated item into a slice about a new flag. The restore is the part that belongs to the flag this PR adds.

**The test pins the premise as well as the fix**, so that stays true:

```python
assert "write_data_report_from_binary(" in converter, (
    "the converter no longer writes the committed data report, so the restore in "
    "update_data.sh's --binaries-only branch is dead and should go"
)
```

When TOOL-6 lands, that assertion fails and the restore gets removed with it. A test that only checked for the restore would have kept it alive for ever.

Verification: `pre-commit run --all-files` clean, `make testall` 3,580 passed.

Issue #502, opened 2026-08-13
First-class `zoneinfo` / UTC-offset helpers


jannikmi:
The library returns a timezone *name*, and essentially every caller's next line converts it to
something usable. `docs/2_use_cases.rst` answers both of the common follow-ups — "make an aware
datetime" and "get the UTC offset" — with a one-line pointer to a script in `examples/`.

Proposal: promote them.

```python
tf.zoneinfo_at(lng=..., lat=...)              # -> zoneinfo.ZoneInfo | None
tf.utc_offset_at(lng=..., lat=..., when=dt)   # -> timedelta | None
tf.localize(dt, lng=..., lat=...)             # -> aware datetime
```

`zoneinfo` is stdlib, so this adds no dependency.

## The trap worth owning

Ocean zones come back as `Etc/GMT+X`, and that naming has an **inverted sign convention**:
`Etc/GMT+5` is UTC**−**5. A user computing an offset from the returned string by hand gets it
backwards, and nothing in the package or the docs warns them. Since the packaged data includes ocean
zones, every coordinate at sea hits this.

That is the strongest argument for owning the conversion rather than leaving it to callers: the
library knows the convention, the caller often does not.

## Scope

- new methods on `AbstractTimezoneFinder` / `TimezoneFinder`, mirrored in `global_functions.py`
- `docs/2_use_cases.rst` and `docs/4_api.rst`
- the `Etc/GMT±X` note belongs in the docs regardless of whether the API lands

A legitimate smaller first step is to put real inline code (and the sign-convention warning) on the
use-cases page and leave the API alone. That captures most of the value for a fraction of the work.

Confidence note: no issue asks for this, which is also the argument *for* it — nobody files an issue
for a line they can write themselves, they just write it, and occasionally get the sign wrong.

jannikmi:
**Decided, 2026-08-21 — ship the full set:** `zoneinfo_at`, `utc_offset_at` and `localize` on `AbstractTimezoneFinder`, mirrored in `global_functions.py`. Chosen over docs-only and over an offset-only helper.

`zoneinfo` is stdlib, so this adds no dependency; what it adds is a permanent public surface. It is additive and so needs no major — but it should ride the same release as the breaking API changes being batched (dropping the unused `in_memory` parameter, and narrowing the package's attribute surface via PEP 562), so the API documentation is rewritten once rather than three times.

[...]

PR #580, opened 2026-09-01
API major (1/2): lazy public surface, drop the inert in_memory, add the zoneinfo helpers

jannikmi:
Three decided register items that must land before the 9.0.0 tag is cut, in the order the sequencing file records as a precondition: API-2, then API-1, then GH-502.

**Based on #577**, not `master` — it restructures the `contributing/` tree this pull request edits. Merge that first.

## Why this is two pull requests

PERF-1 (`is_ocean_timezone` without a regex) and BATCH-1 (the batch form of `timezone_at_land`) are the other two items held for this major. They are a self-contained query-path change with their own measurement argument, and they ship in a follow-up stacked on this branch. Splitting there keeps the review of a breaking public surface from being interleaved with the design review of a new batch API; both land before the tag.

## API-2 — the package's attributes are `__all__` again

`timezonefinder/__init__.py` resolves its public names on first access (PEP 562) instead of importing them eagerly. Importing the finder classes bound `timezonefinder.timezonefinder` as a package attribute and, through it, every module that one imports — so `timezonefinder.utils.validate_coordinates` was as reachable as the seven documented names while carrying no stability promise.

The break is narrow: `import timezonefinder.utils` still works and still binds the submodule. What stops working is reaching one as an attribute of the package without importing it. `import timezonefinder` is also cheaper, since it no longer pulls in `numpy`, `h3` and the polygon readers.

The separable one-liner in the same item ships with it: the stdlib `PackageNotFoundError` the version lookup catches is renamed `_PackageNotFoundError`.

`tests/test_public_api_surface.py` pins this, including in a subprocess — any other test importing a submodule binds it on the package for the rest of the process, so the "binds nothing" half cannot be asserted in-process.

## API-1 — a parameter that cannot do anything now says so

`in_memory` is gone from `AbstractTimezoneFinder.__init__` and from `TimezoneFinderL.__init__`, which was a pure pass-through and is deleted outright. Neither loads polygon data, so there was never an access mode to select. `TimezoneFinderL(in_memory=True)` now raises `TypeError` — which is exactly what `command_line.py` has always answered for `--in-memory` with `-f 3`/`-f 4`, under the comment that accepting it "would promise a speedup it cannot deliver". `TimezoneFinder(in_memory=...)` is unchanged.

One consequence worth flagging: the construction benchmark measured `TimezoneFinderL` twice, differing only in that argument. The two cases now collapse to one node id, `test_initialization[TimezoneFinderL]`, which orphans two metric histories — the same reasoning `scripts/measure_memory.py` already applied to its own configuration list. `docs/benchmark_results_initialization.rst` is regenerated with `make reports` on the tracked configuration (Darwin arm64, Python 3.14.2, clang, no numba), so the other figures move by run-to-run noise as well.

## GH-502 — the `zoneinfo` / UTC-offset helpers

`zoneinfo_at`, `utc_offset_at` and `localize`, on both finder classes and mirrored as global functions. Additive, but riding this major so `docs/4_api.rst` and `docs/2_use_cases.rst` are rewritten once rather than three times.

The argument for owning the conversion is the `Etc/GMT±X` sign convention, which is inverted: `Etc/GMT+5` is UTC-5, and the packaged data covers the oceans, so every coordinate at sea returns one of these. All three resolve the name through the standard library's `zoneinfo`, which applies the convention correctly, rather than parsing it. `tests/test_zoneinfo_helpers.py` asserts the ocean case directly.

`localize` refuses an already-aware datetime with `ValueError` rather than re-labelling it, which would move the instant it denotes.

The Windows `tzdata` requirement is documented at each call site that returns or builds a `ZoneInfo`, per the item's instruction — the package returns IANA names and carries no timezone database.

## Register

API-1, API-2 and GH-502 are deleted — item files and ranking rows — because the work landed. The two lasting decisions are rewritten into `decisions/public-api-compatibility-and-runtime-loading-decisions.md`: the lazy public surface with the options refused, and the rule that an inert parameter says so. The batched-major decision is rewritten to describe the rule rather than the three item handles it named.

BATCH-2's row is unblocked: the comparator compatibility it waited on merged in #566 on 2026-08-31.

**Deliberately not touched**, because the pull requests that cause them already do it in their own diffs: GH-542 and FMT-2 (retired by #574, which deletes both rows and rewrites the FMT-2/GH-513 pairing line) and GH-513's eligibility (#573 already moves it off "blocked by GH-500"). Doing it here would only conflict with them.

## Verification

`uv run pytest tests -m "not slow"` (2254 passed) and the integration selection, plus `uv run pre-commit run --all-files`. `make hook` cannot install over the shared `core.hooksPath` in a worktree.


jannikmi:
@codex review

PR #581, opened 2026-09-01
API major (2/2): the ocean check by prefix, and a batch timezone_at_land

jannikmi:
The remaining two items held for the 9.0.0 major, in their recorded order: PERF-1, then BATCH-1, which is its dependent.

**Stacked on https://github.com/jannikmi/timezonefinder/pull/580**, which carries API-2, API-1 and GH-502 — which in turn is based on #577. Merge in that order.

## Why this is a second pull request rather than one

The judgement call the task left open. The first pull request is a breaking public surface plus a new helper set plus a documentation rewrite; this one is a query-path change and a new batch API with its own design argument. Reviewing a break and a design in one diff means neither gets read on its own terms, and these two items have no coupling to the other three beyond both being held for the same tag. Both land before it.

## PERF-1 — the ocean check without a regex

`is_ocean_timezone` ran `re.match(OCEAN_TIMEZONE_PREFIX, name)` on the result string of every `timezone_at_land` call. `OCEAN_TIMEZONE_PREFIX` is `r"Etc/GMT"` — no regex metacharacters — and `re.match` anchors at the start, so `str.startswith` is *exactly* equivalent. `import re` goes with it; nothing else in the module used it.

Per the recorded decision this ships as a simplification, not a speed-up: the ceiling is ~250 ns, about 6 % of a mixed `timezone_at_land` workload, which is inside the suite's 3–9 % noise floor. No before/after is claimed. `tests/utils_test.py` pins the equivalence by evaluating the old pattern alongside the new call, since it is a claim about the constant rather than about the code.

## BATCH-1 — `timezone_ids_at_land()` / `timezone_names_at_land()`

The asymmetry a user meets immediately: `timezone_at` batched, `timezone_at_land` did not. Both forms ship, matching the existing pair — the id form primary, the name form the convenience on top — on `TimezoneFinder`, `TimezoneFinderL` and as global functions. Same one-keyword-array-per-axis shape; an `(N, 2)` array is rejected as not one-dimensional rather than read by column position.

The ocean check costs nothing per point. `ZoneNames.ocean_flags()` is a boolean array indexed by zone id, built once per instance on first use, so the whole answer array is masked in one indexing operation. That makes the batch cheaper per point than the scalar method rather than merely equal to it — which is the reason BATCH-1 was ranked below PERF-1 instead of being written with a loop.

Two details worth a reviewer's eye:

- The flag array carries one trailing `False`, so `NO_ZONE_ID` (`-1`) reads as "not an ocean zone" by Python's own negative-index rule instead of picking up the last zone's flag. The last zone of the packaged dataset being an ocean zone is exactly when the missing slot would bite; `test_the_ocean_mask_leaves_the_sentinel_alone` pins it.
- It is built lazily and assigned to `self`, which the free-threading finding asks about. It follows `ZoneNames._lookup_array` exactly, including its argument: a race would build an identical array twice and assign the same content, so no lock is warranted.

`test_the_land_batch_drops_exactly_the_ocean_answers` checks the mask against the zone *names* rather than against the scalar method alone — agreement with `timezone_at_land` would also hold if both were wrong the same way.

## Register and decisions

PERF-1 and BATCH-1 are deleted, rows and item files, because the work landed. The sequencing entry that ordered them is rewritten into the lasting rule: a batch form is not worth adding while its last step is per-point string work.

One recorded decision needed correcting rather than deleting. `decisions/query-performance-and-shortcut-index-decisions.md` refused the id-indexed ocean array and said to re-propose it only against an upstream rename of the `Etc/GMT` family. That refusal was about the *scalar* path and still stands there; the batch changed the price rather than the argument, since it holds ids already. The bullet now says which half shipped, which half is still refused, and that the decoupling from zone naming is now had for the batch and not for the scalar method.

## Not in scope

No benchmark is added for the new methods. Measuring the batch lookups in CI is BATCH-2, whose own row this stack unblocks, and adding node ids here would force a second regeneration of the committed reports for a measurement that item is about designing properly.


jannikmi:
@codex review

PR #599, opened 2026-09-03
A release gate that nothing invokes now fails a test

jannikmi:
Reviewing the review findings against the last five merged pull requests, one shape dominates: **a check that cannot run, or runs where it no longer matters.** Every one of those failed open — the pipeline that never consulted the gate looks exactly like a pipeline the gate approved. The rest of the findings were crashes, which are loud and cheap.

The worst instance: `scripts/changelog_fragments.py` shipped in #589 reachable only from a manual `make` target. A release could publish with a fragment left behind, and because `changelog.d/` is pruned from the distribution the bullet would be absent from `CHANGELOG.rst` *and* the package — the release would simply read as if that change never happened. It was green, because the suite exercised the check itself and nothing asserted that the release path reaches it.

## What changed

**`tests/test_release_workflows.py` now holds a table** mapping every command under `scripts/` to the irreversible step it must precede, and fails on a command with no row. "Nobody wired it up" stops being a state reachable by forgetting: adding a command forces the question, and answering it with a release gate *creates* the ordering assertion, which was previously written by hand once per gate. The two hand-written ordering tests collapse into one parametrised over the table, each carrying what a silent pass would cost, reported in the assertion. A gate whose boundary is not `build.yml`'s names the test that pins it, and that pointer is asserted to resolve, so the classification cannot rot into a claim nothing checks.

**Contributor memory** states the fail-open class and the independent review that catches it, in the routed testing rules — a pass's own final gate runs only the tests its author wrote, so it cannot fail on the invariant its author did not think of. The improvement-pass workflow routes to that review before the pull request is opened, so the automated round is spent on what an independent reader could not already see, and a confirmed finding that would fail silently is never deferred to a follow-up item.

## Verification

`make hook` clean; 1705 unit tests pass on this rebased head.

Mutation-checked, each failing its own assertion:

- removing the changelog-fragment gate step from `build.yml` → `jobs in build.yml that publish without \`scripts.changelog_fragments\` having already run: ['publish-pypi', 'release']`
- adding an unclassified `scripts/` command → `new \`scripts/\` commands with no row in SCRIPT_GATES`
- renaming `test_a_draft_update_is_never_merged` → `scripts/data_update_guard.py's ordering ... is pinned by nothing`

## Judgment calls

The classification is exhaustive over commands rather than over all of `scripts/`: a gate is always invocable, and library modules would have been 11 rows of noise. It would have fired on #589 — the pass added `scripts/changelog_fragments.py`, the row would have been demanded, and the item's own acceptance criteria already said the answer was "before the release publishes".

Substance went into `contributing/development/testing-strategy-and-change-scope.md` rather than the workflow file, which is at 1950 of its 2000-word budget; the workflow keeps one routed sentence.


jannikmi:
Second commit, from analysing what the **maintainer** asked for after each pull request opened — as distinct from what the automated review asked for. Three of the four interventions were not about code correctness at all.

| PR | Post-open commits | Who drove them |
|---|---|---|
| #588 | 0 | merged 8 min after opening, tox job still running |
| #589 | 3 | Codex — but only because the maintainer re-triggered it three times |
| #590 | 6 | **the whole PR is maintainer-requested**, plus 1 Codex fix |
| #591 | 0 | merged 8 min after opening |
| #592 | 3 | 2 Codex, then a maintainer design correction 4.5 h later |

**#592 is the important one.** The pass built the guardrail as an exit from `update_data.sh`, stopping the preparation. The recorded decision says a tripped guardrail blocks the *auto-merge*, and names the mechanism: "a guardrail is a job in `build.yml` — no new blocking machinery". Two Codex rounds passed it. Only the maintainer had read the decision. Worse, the cascade: a Codex P2 ("persist the rejected diff") produced an artifact upload, which broke the wheel hand-off test, which was then narrowed — and `9f58272` threw all three away once the seam was corrected. The finding was a symptom; fixing it where it was reported patched over the design error.

So the independent review from the first commit now gets **any recorded decision the change is bound by** — the one kind of intent worth handing a reviewer, since no test fails on correct code implementing the wrong decision.

**#589's five defects** were found by three *maintainer-requested* re-triggers, each round reading code the previous round had not. The workflow rule landed in #590; the merge-side check was missing. The merge round now compares the commit a review names against the head: a pull request whose fixes postdate every review of it has had its riskiest code read by nobody. That also covers #588 and #591, whose zero findings mean nobody looked, not that they were clean.

**#590's own body** excused `--no-verify` with "check-manifest takes minutes" — wrong by two orders of magnitude, and self-corrected in the same body. A wrong reason costs more than a missing one because it retires the question, so an empirical claim in a body or commit message is now stated where it was measured and marked where it was not.

`make hook` clean; 1705 unit tests pass. No new mechanism here — this commit is prose only, and deliberately so: decision compliance is not mechanically checkable, and claiming otherwise would be the same mistake as a guard that fails open.

jannikmi:
Two more commits: the symptom question you asked for, and five older merged pull requests read the same way.

## `8e9d40e` — ask what a finding is a symptom of

The pass now asks it before fixing a finding where it was reported, and the testing rules carry the reasoning next to the observability ranking they already state. The routed sentence also names the recorded decisions the reviewer is handed, so it matches the rule it routes to.

## `1214b63` — five older instances

| PR | What happened after it opened |
|---|---|
| [#577](https://github.com/jannikmi/timezonefinder/pull/577) | Codex asked for a withdrawn register item to be restored. Reply: *"the user explicitly requested removal of the redundant item, so I clarified the register rule instead of restoring it."* Four further commits threaded that exception through the register rules, the pass workflow and a discovery surface — three of them prompted by findings that the rule now disagreed with itself |
| [#564](https://github.com/jannikmi/timezonefinder/pull/564) | Codex **P1** *"Split this batch into single-item improvement passes"*, answered by making batching the policy: *"Keeping this bounded batch is the intended new policy"* |
| [#577](https://github.com/jannikmi/timezonefinder/pull/577) | *"Stop automatic Codex review retriggers."* Five days later that was the whole of #589's gap — every review saw only the opening commit, and three maintainer-requested re-triggers found five defects |
| [#569](https://github.com/jannikmi/timezonefinder/pull/569) | Four review rounds widening one boundary: retry the release API → re-unpack the verified archive → retry urllib transport exceptions → retry raw socket resets. The reply admits it: *"closing the gap from the download side last round left it open from this one."* The last commit is named for the boundary it should have drawn once |
| [#586](https://github.com/jannikmi/timezonefinder/pull/586) | Merging master into the 9.0.0 branch kept #588's `build.yml` change but resolved the changelog hunk in favour of the release branch. The section would have frozen *"while silently omitting a change to what `pip install` resolves on Alpine."* Caught minutes before publication |

Three more of the same families, not counted above: [#578](https://github.com/jannikmi/timezonefinder/pull/578) (`--binaries-only` promised no report writes and left a tracked file dirty), [#582](https://github.com/jannikmi/timezonefinder/pull/582) (the noise gate read off the numba kernel while the committed pages assert clang), [#573](https://github.com/jannikmi/timezonefinder/pull/573) (`6932cb6` deletes 106 lines from `command_line.py` under an empty commit body).

## Rules added

- **A finding is never retired by amending the rule it cites, inside the pull request it landed on.** Two of the five did exactly this. It is the most dangerous shape in the set because it is invisible in the diff: the change passes because the standard moved. Either the rule change is its own deliverable, argued on its own merits, or the finding is fixed. Not a claim that #577's deletion was wrong — you asked for it — only that the amendment had to be its own argued change.
- **A rule that removes a check is code that removes a check**, and states what now passes unexamined. #577's retrigger ban fixed a nuisance and opened the hole that cost #589 five defects, with nothing connecting the two.
- **The second finding on one mechanism is the signal, not the fourth.** #569 is the older, clearer version of the #592 cascade.
- **A change made because the maintainer asked says so in its commit.** This was the hard part of the analysis: your feedback arrives in conversation and leaves no trace on the pull request, so I could only attribute the cases where a body or a reply volunteered it. An intervention nobody records cannot be counted or learned from — the same fail-open shape as the guards.

#586 is worth noting as a class already closed: one fragment per change means that merge conflict cannot happen again, which is the clearest evidence in the set that #589's work retired a real failure mode rather than a hypothetical one.

`make hook` clean; 1705 unit tests pass.

Issue #497, opened 2026-08-13
Profile where per-query time actually goes (blocks the batch API, #301 and #364)


jannikmi:
Several proposed optimisations are each justified by an assumption about which stage of a lookup
dominates, and those assumptions are not all consistent with each other. One measurement settles
them. #477 already showed that one such assumption — that a dict lookup is negligible — is wrong in
the direction that matters, so this is not a formality.

## What is already known (#477, numba backend, macOS arm64)

| query | total |
|---|---|
| unique-zone hit | 0.94 µs |
| ocean hit | 0.96 µs |
| NYC | 14.9 µs |
| near a border (Basel) | 28.9 µs |
| — of which `shortcut_mapping.get` | 68 ns |

So the unique-zone query is ~1 µs with no geometry touched, and the ambiguous query is 15–30× that
and clearly dominated by the polygon loop.

## What is missing

1. **The breakdown inside those totals.** Attribute time to `validate_coordinates`,
   `h3.latlng_to_cell`, shortcut resolution, `zone_name_from_id`, bbox rejection,
   `ffi.from_buffer` marshalling, and the point-in-polygon kernel itself.
2. **The clang-without-numba backend.** #477 measured numba only. `utils.py` binds the
   implementation at import time and numba wins whenever it is importable, so a local `uv run`
   never exercises the C path — *which is the configuration a plain `pip install timezonefinder`
   produces*, and the one CI tracks. The per-polygon `ffi.from_buffer` marshalling cost that
   motivates the native-candidate-loop idea **does not exist on the numba path at all**, so
   profiling only numba would make that idea invisible.

## How

- `line-profiler` is already a `proto` dependency; `scripts/shortcuts.py` has the `profile`
  decorator fallback idiom to copy.
- Do it in `prototypes/`, alongside the existing single-resolution and BFS studies. Not shipped,
  not a benchmark, no thresholds, no CI, no committed figures.
- Use the committed deterministic fixtures in `tests/fixtures/benchmarks/` rather than sampling
  fresh points — they already carry strata, and two runs of the same commit then execute the same
  workload.
- `scripts/assert_acceleration_path.py` exists to pin which backend is active; use it so the
  results are labelled correctly.

**Report per stratum × per backend, and paste the table back into this issue.**

## What it decides

- **Batch API** — whether it is amortising real overhead or shaving noise.
- **#477** — whether the flat-array layout needs to keep a separate scalar path.
[...]

PR #600, opened 2026-09-03
Compare all three point-in-polygon paths, and stop the docs claiming which is fastest

jannikmi:
Stacked on #593 — review that first; this branch targets it, not `master`.

The benchmark pages describe one configuration (clang without Numba, what a plain `pip install` gives you). That is right for a trend chart, and it left a reader unable to ask what installing Numba buys them — while the hand-written docs answered that question, wrongly, in four places.

## The ranking now lives in exactly one place

`docs/benchmark_results_acceleration_paths.rst` measures all three paths against each other and is regenerated by `make acceleration-paths` with the rest of the reports. Everything else points there.

That is the point of the change, not a side effect. A ranking in prose goes stale silently, and this repository had already proved it: `~400x` for pure Python appeared in `README.rst`, `docs/7_performance.rst` and twice in `docs/architecture.rst` and matched nothing measurable — the packed kernel puts the gap at 4x on small rings. `7_performance.rst` separately read the dispatch rule ("Numba takes precedence") as a speed claim, which the repo's own #497 decision record already contradicted.

What the hand-maintained docs keep is what does not rot: the dispatch rule, the contract that all three return identical answers, and the caution that being *preferred* says nothing about being *faster*. The same treatment is applied to generated prose and source comments, which are no more self-updating — the polygon and timezone-finding pages' cross-link names no direction, and the ratios in `scripts/assert_acceleration_path.py` are gone.

## Why two environments, and why one ratio is missing

Numba and pure Python are the same source decorated or not, so **no process holds both**. Each pair is compared in its own process against the C extension both environments hold, which keeps everything `benchmarks/candidate_comparison.py` exists to hold: alternating order, one shared draw per round, and the round-level win count.

The Numba-against-pure-Python ratio would have to cross that boundary, so it is **not computed**. Across processes only the best-round ratio survives — the single estimator this repository has already been misled by (13.3 % vs 0.3 %, per the methodology page).

The two runs' shared clang baseline is published beside the tables as the reader's comparability check rather than used as a gate, and it turned up something worth publishing: **installing Numba changes more than the kernel**. `validate_coordinates` then calls two `njit`'d scalar helpers rather than two plain comparisons, and every query pays that before any geometry — so the divergence is largest on the unique-shortcut rows, where validation is most of the query.

## The enabling library change

`PolygonArray` wrapped its payload for the bound backend at construction but read `utils.inside_polygon_packed` as a module global per lookup. Only a comment held the pair together, and two finders on different paths could not coexist — which is what a paired A/B through the public API needs. The kernel is now captured beside the buffers.

Measured on the query path it touches (`benchmark_core`, Apple M1 Pro, `min`, head–base–head sandwich): every row unchanged, the two head passes agreeing to within 0.2 %, far inside the 3–9 % same-machine floor.

## Notes for review

- `docs/benchmark_results_acceleration_paths.rst` joins `REPORT_FILES` in `scripts/benchmark_report_artifact.py` — without that the release's completeness check would fail on it.
- No pytest-benchmark node ids are added, so the trend chart and the per-PR job are untouched. `make acceleration-paths` rides `make reports`.
- The polygon page's `*_python` rows are relabelled after the path that produced them; the node id stays put, since it is the chart's join key.
- `decisions/query-performance-and-shortcut-index-decisions.md` recorded the two kernels as "the same speed". Its evidence stands — it measured the *bare* kernel, before polygon layout 3 — so the entry keeps it and is extended rather than rewritten.
- `make reports` ends with plain `uv run` steps, which prune Numba from a dev `.venv`. Pre-existing, hit while doing this; `uv sync --all-groups` restores it.

## Verification

- `uv run pre-commit run --all-files` clean; `make test` and the full `pytest` suite pass (3,775 tests).
- New: two finders on different paths coexist and agree over 1,000 ambiguous points — fails on the previous code.
- New renderer tests cover the machine-identity refusal, the one-run-per-environment refusal, and the published baseline spread.
- `cd docs && make html` builds with no warnings.


jannikmi:
Superseded by #603.

This was not closed on its merits: it was stacked on #593, and deleting that branch when it squash-merged closed this PR one second later. Reopening it in place would not have worked either — #593 was squash-merged, so master carries its work as `bb28613` and none of its three commits are ancestors of master; retargeting this PR to master would have shown #593's work as a duplicated diff.

#603 is the same work rebased onto master, with those three commits dropped. Diff falls from 43 files / 2,283 insertions to 31 / 1,840.

PR #619, opened 2026-09-05
Exempt an entry-retiring pull request from the independent review

jannikmi:
## What changed

A pass that **removes** an item instead of implementing it no longer owes the pre-open independent review. Requested by the maintainer after #613.

The rule is stated once, in `contributing/development/testing-strategy-and-change-scope.md`, which owns the review requirement. The improvement-pass workflow, the review-settlement rules and the merge round's park list link to it rather than restating it — the merge-round and improvement-pass skills are pointer stubs, so "the skill" is the workflow file each routes to.

## What is exempt, and what is deliberately not

**Exempt:** the whole diff is contributor-memory files, the subject is retiring one entry — deleting its item file and ranking row, or moving that row to `Closed` — and every line added is either mechanical or text already reviewed on a merged pull request.

**Not exempt, and this is where I did not implement the request literally:**

- **A retirement that *writes* a refusal or a finding.** A `decisions/` or `checked-and-found-sound/` record is a standing instruction to every later pass: each opens with "do not re-propose" or "do not re-raise", the register keeps it *never deleted*, and the bullet directly above this one in the same file calls a recorded decision "a repository invariant and the one kind of intent worth handing over" to a reviewer. Exempting that would have excused the longest-lived memory a pass can write, on the grounds that nothing was being written.
- **Restating another entry's blocker, sequencing or measurement.** The register rules require greping the retired ID across `contributing/improvements/` and rewriting what referenced it. Deleting a dangling handle is mechanical; rewriting a blocker changes what the next pass takes first, so it is read.
- **A refinement into slices**, which creates entries and states counts, blockers and sequencing a reader can check — the review of #614 found four of those wrong.
- Anything touching source, tests, configuration, or published documentation.

## The cost, stated at the rule

The contract requires a rule that removes a check to say what now passes unexamined, so it does, and names the real incident: **the review of #613 — itself a removal — found its retained evidence naming the wrong mechanism for a precondition.** That is the case that shaped the boundary above rather than an argument the boundary ignores.

What is left inside the exemption is the claim that the entry is finished. Nothing fails on a confidently wrong memory, so that claim gets an artefact instead of a reviewer: **an exempt pull request names, in its body, the merged pull request whose work retired the entry, or states that nothing was retained and the item was wrong.** The merge round parks one that does not carry that sentence, so skipping it is not a state reachable by forgetting — the same shape `tests/test_release_workflows.py` uses, and what the "a guard that never runs" bullet two entries earlier demands.

## Behaviour impact

None on the package. The effect is on process: a removal-only pull request opens without a pre-open review, and the merge round stops parking it for a missing one — but starts parking it for a missing receipt. Everything after opening is unchanged: an exempt pull request is still opened, watched, and settled like any other, and a finding that arrives on one is still answered.

## Judgment calls

- **The exemption is narrower than "a pass removing an item".** Retiring an entry while authoring new standing memory is common — #613 did exactly that — and it is the sub-case the evidence says is riskiest. Widening it to cover that is a one-line edit if the maintainer wants it; I have not made that call, and this section is where to overrule me.
- **A `Closed`-move is inside the exemption when it writes nothing new**, even though the entry's text survives. What a reader would check is the new prose, and there is none.
- **The receipt is a sentence in the body, not a test.** The pull-request body is not a file a test can read, and the merge round already parks on body-level facts, so that is where the enforcement went.

## Verification

- `make hook` — clean.
- `make test` — 1,967 passed, 1,893 deselected.
- `uv run pytest tests/test_contributor_memory.py tests/test_agent_skills.py tests/test_improvement_ledger.py` — 14 passed; word budgets hold, all four files under the 2,000-word canonical limit (1,678 / 533 / 1,823 / 1,870).
- No changelog fragment: the diff is entirely `contributing/`, which the changelog policy exempts outright.
- Diff is four contributor-memory files. No source, no test, no packaged data.

## Independent review

This change is not itself a removal, so it took the review it is about. Reviewed on `d2c946e` by a reader given the diff and the routed rules but not the request or the reasoning. Four findings, all confirmed and all fixed on the head:

1. **The stated premise was false for the rule's own second branch.** It justified itself with "ships no instruction" while covering evidence written into `decisions/` — which is an instruction, by that file's own first line. Fixed by putting new prose outside the exemption.
2. **The boundary did not exclude "retires an entry *and* something else"**, while the register rules *mandate* the something else. Fixed by splitting mechanical deletion from restatement.
3. **The rule was duplicated rather than routed, and the three copies already disagreed** — a flat exemption in the testing rules, "pre-open only" in the improvement pass, "everything still applies" in the settlement rules. Fixed: one statement, three links.
4. **The self-check had no teeth**, and the same file two bullets earlier demands a form where the omission cannot happen quietly. It also removed the merge round's park without putting anything back. Fixed by the receipt above.

The reviewer's closing suggestion — restrict to entries whose work demonstrably shipped, no new `decisions/` prose, no reference rewrites, state once, give the check a form the merge round can park on — is what the head implements.


jannikmi:
Closing this unmerged: **the evidence from this very session is that removal-shaped pull requests did benefit from review.**

The exemption was proposed on the reasoning that a removal ships no behaviour, so an independent reader has nothing to check the repository against. Three pull requests opened in the same session say otherwise:

- **#613 was a pure removal, and its review found a real error** — the evidence it carried into `checked-and-found-sound/runtime-geometry-and-data-checks.md` named `pip_with_bbox_check` as the only route to `PolygonArray.pip`, when the path carrying most lookups reaches it through `TimezoneFinder.inside_of_polygon`'s own `outside_bbox`. That is a wrong fact in a record whose entire purpose is to stop a later pass re-deriving it, and nothing in the suite fails on it.
- **#614 shipped no code either, and its review returned four confirmed findings**, including line counts inflated by a heuristic that counted directive bodies as prose.

So the class this exemption was drawn around is the class the session's own evidence says is worth reading — not because a removal is risky to run, but because a removal's whole content is an assertion about the current code, and a wrong assertion there is invisible.

The work is on `claude/review-exception-for-removals` if the question is reopened. What it established and what a future proposal should answer:

- A `decisions/` or `checked-and-found-sound/` record is a standing instruction — each opens with "do not re-propose" or "do not re-raise", and the register keeps it *never deleted*. Retiring an entry usually means writing one, so "a removal writes nothing" is false for the common case.
- Retiring an entry also rewrites whatever referenced the ID, which the register rules require; a restated blocker changes what the next pass takes first.
- Whatever replaces a review has to be something a later reader can see. A self-check leaves a pass that did it and a pass that skipped it with byte-identical diffs.

Issue #524, opened 2026-08-18
Move timezonefinder under packages/ for a symmetric workspace layout


jannikmi:
Follow-up to #446. Its first slice makes this repository a `uv` workspace with **one** member
under `packages/` while `timezonefinder` itself stays at the repository root:

```
pyproject.toml                 # timezonefinder (workspace root AND a member)
setup.py
MANIFEST.in
timezonefinder/
packages/timezonefinder-data/  # the only thing under packages/
scripts/  tests/  benchmarks/  docs/
```

That asymmetry was deliberate — it keeps the split off the release path — but it is a compromise,
and this issue tracks making the layout symmetric:

```
packages/timezonefinder/
packages/timezonefinder-data/
```

## Why it was not done in #446

Everything that builds and publishes the code package is root-relative, and all of it currently
works:

- `setup.py` names `cffi_modules=["timezonefinder/build.py:ffibuilder"]`
- `[tool.cibuildwheel.*]` lives in the root `pyproject.toml`, and `build.yml`'s `make-wheels` job
  runs cibuildwheel against the repository root — a move needs `package-dir` throughout
- `MANIFEST.in` is written in root-relative terms (`recursive-include timezonefinder …`,
  `graft tests`), and the `check-manifest` hook's ignore list
  (`.pre-commit-config.yaml`) is root-relative too
- `scripts/configs.py` derives `PROJECT_ROOT`, `DATA_VERSION_FILE` and `PYPROJECT_FILE` from its
  own location
- `tox.ini`, `readthedocs.yml` and `docs/conf.py`'s `sys.path` insertion all assume the current
  shape

So the move is a wide, low-insight diff across exactly the machinery whose failure mode is "no
release goes out", and #446 had enough surface already.

## What it would buy

- **Neither distribution is privileged.** Today the root `pyproject.toml` is simultaneously the
  workspace root and one member's manifest, which is legal but reads as an accident.
- **A third distribution stops being a special case.** This is the real trigger: #332 (the reduced
  `timezones-now` dataset) becomes a second data distribution, and adding it to a symmetric
  `packages/*` is mechanical where adding it to the current shape means deciding all over again
  what lives where.
- **`scripts/`, `tests/`, `benchmarks/` and `docs/` stay at the root either way** — they serve
  every package, so the layout does not become uniformly flat, only unambiguous about which
  directories are distributions.

## Open questions

- Whether to do it at all before #332 needs it. Absent a third distribution this is close to
[...]

Issue #362, opened 2025-10-17
Code Quality: reuse the PolygonArray binaries in the file conversion process


jannikmi:
TODOs:
- [ ] Write the polygon base binary files for boundary and hole polygons as early as possible in the file conversion script
- [ ] use the PolygonArray class afterwards
- [ ] if useful, create a specialised child class of PolygonArray for the file conversion script and attach required util functionality directly to the polygon data. keep the util functions in the functional programming style as much as possible
- [ ] cleanup the uneccesary file conversion helper util code (due to simplification)

This would improve the code quality by reducing code duplication

jannikmi:
Closing: the duplication this issue is about is gone, but by a different route than the first two TODOs describe, and those two will not be done.

**What shipped** — #676, merged as 10bbbbf. The converter's boundary and hole rings are reached through one collection object each, named and indexed as the runtime names and indexes the same two: `TimezoneData.boundaries` / `.holes` against `TimezoneFinder.boundaries` / `.holes`, with `coords_of` and `ids_of` as methods, `nr_vertices` as a column, and `holes_of_poly` as the finder spells it. `TimezoneData` no longer exposes the eight index-parallel per-polygon structures (`polygons`, `polygon_lengths`, `poly_boundaries`, `hole_boundaries`, `holes`, `all_hole_lengths`, `polynrs_of_holes`, `original_polygons`) that callers had to keep in step by hand. That is the third TODO, which is where the code-quality win actually was: a reader of either side can now read the other.

**Why the first two TODOs are refused.** Writing the binaries early and reading them back through `PolygonArray` costs a payload decode on every access. Measured over the packaged boundaries (1,322 rings, 7,925,313 vertices, `in_memory=True`): `coords_of` is **193 µs per call against 0.04 µs** for indexing the in-memory ring. `Hex.lies_in_cell` reads a ring once per (cell, candidate) pair while compiling the shortcut index, so this is not a slower parse but an unusable one. Materialising every ring once into memory instead is affordable (~0.26 s, ~63 MB) but buys nothing on its own, since the producers still have to run. The decode is exactly what the packed polygon layout exists to charge `get_geometry()` for; the converter is the caller that must not pay it.

So the converter shares the runtime's *vocabulary*, not its storage. That refusal and its measurement now live at the accessor itself (`PolygonCollection.coords_of`) and as a rule in `contributing/development/coding-design-and-state-management-rules.md`, so the next contributor meets it where the code is rather than re-deriving it.

[...]

Issue #332, opened 2025-08-07
use reduced timezone dataset


jannikmi:
Usage of the [timezones-now dataset](https://github.com/evansiroky/timezone-boundary-builder?tab=readme-ov-file#same-since-now) with 92 instead of 444 timezones
Discussion: https://github.com/jannikmi/timezonefinder/discussions/323
Draft PR: https://github.com/jannikmi/timezonefinder/pull/324

jannikmi:
Closes by PR #324

jannikmi:
The new release will revert back to the full dataset due to the discussions in #363

jannikmi:
This becomes much easier after #446: rather than a build-time switch inside one package, the
reduced `timezones-now` dataset is simply a second published data distribution that a user installs
instead of the default. That removes the hard part — making one package behave as two datasets —
and turns it into a packaging decision.

Suggest designing it as a deliverable of #446 rather than ahead of it.

Context: this comes out of a wide-angle review of the package; #506 is the roadmap that ranks it against everything else open.

jannikmi:
**Decided, 2026-08-21 — parked until the official mapping exists.**

Shipping now with a hand-maintained mapping was declined, and so was closing this outright. The `timezonefinder-data` split removed the hard engineering problem exactly as the earlier comment predicted — a workspace member, a `DATA_DIR` indirection and a version scheme all exist — so this is worth doing *when* #334 unblocks.

What stops it is the mapping itself. Until evansiroky/timezone-boundary-builder#195 lands there is no official 444 → 92 table, and a hand-curated one is the same liability the zone-precedence engine was rejected for: timezone answers depending on a table somebody maintains by hand, in a package whose selling point is that it does not simplify.

Two costs worth stating for whoever picks this up, since they are not in the body above: a second published data distribution means a second release axis and a second ~52 MB PyPI project; and **every zone name in the documentation becomes dataset-dependent**. That is not hypothetical — the reduced dataset's `Europe/Paris` answer for Berlin coordinates once annotated the default dataset's running example in eleven places across `README.rst` and `docs/1_usage.rst` before it was corrected wholesale.

#334 already tracks the upstream trigger, so nothing else needs to watch for it.

Related: the hand-derived 18-pair fragment of this mapping that had been sitting unused in `tests/locations.py` since #324 is being deleted. Whatever this issue eventually uses, it is the official table, not that.

jannikmi:
[...]

PR #630, opened 2026-09-05
TOOL-2: enforce B023, and refine the lint item into slices

jannikmi:
## What changed

`B023` — a function defined inside a loop that closes over a name the loop rebinds — is enforced. It was selected as part of `B` and then ignored again, so nothing in the tree was held to it.

All 24 sites were read individually, and all 24 are false positives: every closure is defined and fully consumed inside the iteration that makes it. Rather than bind up to seven loop variables as default arguments to say so, the three `for polygon_dir in (...)` bodies in `timezonefinder/_data_integrity.py` move into module-level `_validate_*_of` functions — the closures stay exactly where they are, with no loop around them — and `at_target` in `tests/test_border_sampling.py` becomes a module-level helper taking its distance explicitly.

**No behaviour changes.** The extraction is re-indentation plus a `def` and a call; verified by diffing the normalised token streams of the old and new module, which differ only by the new signatures, the two calls, and two `ruff format` reflows the dedent freed up.

`tests/test_lint_configuration.py` gains `SELECTED_RULE_PROBES`, a probe set for the rules `select` adds beyond ruff's defaults (`B023`, `RUF013`), asserted the same way the file already asserts the default rules: by linting code that breaks each rule. `select` and `ignore` are separate lists, so a name in `select` says nothing about whether the rule reaches any code — returning `B023` to `ignore` would now fail.

## This is a refinement, and this is its first slice

`TOOL-1` held six ruff rule families plus the 0.16 version bump: 73 sites to judge across disjoint rules, and a dependency pin. That is not one reviewable pull request, so it is refined into slices that each stand alone under one changelog sentence. It is deleted with its ranking row; nothing narrows its ID.

| slice | what | state |
|---|---|---|
| **TOOL-2** | `B023`, 24 loop-closure sites | **shipped here**, and retired in this branch's third commit |
| TOOL-3 | `B905`, 27 `zip` sites needing a per-site length judgement | open |
| TOOL-4 | `A001` / `A002` / `PLW2901`, 7 sites — adopt or record the refusal | open |
| TOOL-5 | `PLR09xx` complexity limits | **needs a decision** — refuse, ratchet above the tree, or adopt ruff's defaults |
| TOOL-6 | raise ruff off the 0.15 line | **needs a decision** — it is a dependency pin plus the lockfile, which a pass may not change |

TOOL-6 is the one worth flagging: the work is real and priced, but the improvement-pass boundary forbids a pass from touching a dependency or the lockfile, so it can never be taken by one as it stands. The entry says so rather than sitting `open` for a pass that would have to refuse it.

Every count was re-measured on ruff 0.15.22 rather than carried over. Two had drifted: `PLR0913` is 7 where TOOL-1 recorded 10, and `A002` is 2 where it recorded 1.

## Why this item, and what was skipped above it

| row | why not taken |
|---|---|
| GH-364 | blocked on an h3 release |
| GH-332 | parked |
| PROF-1 | [#628](https://github.com/jannikmi/timezonefinder/pull/628) is open on it |
| **PERF-2** | **claimed and then yielded — see below** |
| PERF-8 | [#625](https://github.com/jannikmi/timezonefinder/pull/625) is open on it |
| GH-524 | needs a maintainer decision |
| GH-362, PERF-6 | both in [#609](https://github.com/jannikmi/timezonefinder/pull/609) |
| FMT-3 | [#626](https://github.com/jannikmi/timezonefinder/pull/626) is open on it |

**PERF-2 was claimed, implemented, and then reverted.** It ranked first among the free rows and its claim ref was clean, so this session took it and wrote the change — including a seam test pinning `_zone_id_among`'s two zone-id reads against the candidate list's own ids. Reading #628's body while gathering the profiler context revealed that #609 already implements exactly PERF-2's re-scoped form, down to the same signature changes, and that #628 had already yielded to it for the same reason. Racing an open pull request in one file is what the workflow forbids, so the implementation was discarded, the claim ref released, and the branch deleted. Nothing of PERF-2 is in this diff. `#609`'s own register commit retires it.

The cost of that detour was that the claim ref and the branch existed for a few minutes; the lesson — that an open pull request's *body* can be the only place a ranked item's fate is recorded, so the re-verification sweep has to read the open pull requests and not only their titles — is not new memory, it is the workflow's existing "inspect whatever is associated with an item before selecting it", read too narrowly.

## Judgement calls

- **Extraction over default-argument binding.** `check` in `validate_block_payload` closes over nine names; binding them as defaults would present nine required inputs as optional and put the rule's workaround in the signature. Extracting the loop body removes the loop instead, which is what the rule is actually about, and each of the three directory checks is now independently readable.
- **The probe test covers `RUF013` too**, not only the rule this change enables. It is the same failure mode and the same two lines, and a probe set with one entry invites the next rule to be added without one.

## Verification

- `uv run pre-commit run --all-files` — clean. (`make hook` cannot install over this worktree's shared `core.hooksPath`.)
- `make test` — 1986 passed.
- `make testall` — 3883 passed.
- `uv run ruff check --select B023 .` — no findings, where `origin/master` has 24.
- Diff contains only intended paths; no packaged data touched, and the working tree is clean.
- Rebased onto `origin/master` (`dceaff6`) before the gate; the base has not moved since.

## Independent review

Reviewed at high effort against the diff and the routed rules, with no findings. What was checked beyond the mechanical equality above: each extracted helper still binds every local its body reads (`coordinate_path` in particular, which is derived per directory in all three); and the module-global monkeypatches the existing tests rely on — `PolygonArray`, `HoleArray`, `derive_payload_offset_table` — still take effect, because the helpers resolve those names from the same module's globals at call time.

## Next eligible item

**TOOL-3** — `B905`, 27 `zip` sites. Every row above it is blocked, parked, in an open pull request, or waiting on a maintainer decision.

## Claims

`improvement-claims/TOOL-1` and `improvement-claims/TOOL-2` are released with this pull request, which becomes the durable claim. `improvement-claims/PERF-2` was acquired and released when that item was yielded.


jannikmi:
@codex review

Issue #301, opened 2025-07-17
performance optimisation: sort polygons in shortcuts according to area


jannikmi:
currently the shortcuts contain a list of polygon boundary ids sorted after the total size of the polygons of each zone.

it would be more performant to sort them after the actual area each boundary overlaps with the shortcut. assuming uniformly random distributed query points, the covered area corresponds to the probability of a query point lying within that polygon. use utility libraries (just installed for development) for this area computation.
- consider `shapely`

not-wish:
Hi! I'm looking into this issue and tracing the shortcut generation pipeline. The issue mentions using utility libraries for the overlap area computation, but I couldn't find any geometry library in the current `pyproject.toml` or `codebase`. Am I missing a branch or dependency, or is adding one as part of this issue expected?

jannikmi:
Thanks for looking into this!
Part of working on this, would be to evaluate libraries for computing polygon overlap. [`shapely`](https://shapely.readthedocs.io/en/stable/) for example might support this.
This additional dependency would only be required for compiling new binaries (releases) and not the client user code (query side). This dependency would hence not be an additional dependency for the downstream users. Due to this, it is acceptable if this dependency is large or has other large sub-dependencies etc..

jannikmi:
Closing this as **not worth doing**. The mechanism is sound and the sort key is the right one — overlap area really is the probability a uniform query point falls in that polygon — but the win it can produce has now been bounded, and it is too small to justify the change or the `shapely` dependency.

## What ordering can actually win

The argument for this issue is that a better order rules zones out sooner, so fewer polygons get a point-in-polygon test. That is a **count**, so it can be measured exactly by enumeration over the packaged shortcut index — no benchmark, no machine, no timing noise. Counting the candidates each ambiguous cell tests when no early hit occurs:

| | tests | mean per ambiguous cell |
|---|---:|---:|
| current ordering | 12,600 | 1.199 |
| the best any ordering could achieve | 12,234 | 1.164 |
| **headroom** | **366 (2.90 %)** | |

Only **259 cells** — 2.46 % of the 10,511 ambiguous ones, **0.63 % of all 41,162** — would test fewer polygons under *any* reordering. On the other 99.4 % the order changes which polygon is tested, never how many.

## Why the headroom is that small

Two properties of the current design have already taken most of it:

1. `timezone_at` breaks at `last_zone_change_idx` and returns the final zone **without testing it**, so the last zone in a cell is free.
[...]

PR #638, opened 2026-09-06
GH-301: close the shortcut candidate-ordering question by construction

jannikmi:
## What changed

Register only. GH-301 — re-ordering a shortcut cell's candidate polygons by the area each overlaps the cell — was **built, measured and refused**, along with the rest of its key family and the provably optimal ordering itself. The entry moves to `Closed` and is retained as the measurement record; the [shortcut decisions](contributing/improvements/decisions/query-performance-and-shortcut-index-decisions.md) carry the refusal in one paragraph; and the two rules that read a count of point-in-polygon tests as a workload share are corrected, with this item as their counterexample.

That decisions file sat at 1,992 of its 2,000-word budget, so the H3-resolution record — the largest entry and a bounded subject of its own — is split out into `shortcut-index-resolution-decisions.md`, with a pointer left behind. Its text is unchanged, and every existing link to the parent file, including the two from package source, still resolves.

No package code, no generator, no packaged data, no dependency and no lockfile entry changed. `shapely` was added to the `data` group to take the measurement and removed again.

## Why this item, ahead of the ranking

Directed by the maintainer, together with the decision the entry was missing: a `shapely`-class dependency is acceptable at conversion time, not at runtime. GH-362 outranks GH-301 and was skipped for that instruction; PROF-1 is in flight as #633 and PERF-4 carries a live claim ref, so neither was available anyway.

## The measurement

The entry demanded its enumeration be re-run before it was taken, because a change to the packaged index moves the bound. Eight keys were compiled into real shortcut binaries from the verified 2026c source, spanning the whole per-cell blend of the two properties available — a zone's vertex count in the cell, and the area it covers of the cell. Counts come from wrapping `inside_of_polygon` over the committed fixtures; timings from `benchmarks/candidate_comparison.py`, 61 rounds × 2,500 points, order alternated, one process holding both indices, mapped access, ranked C-extension column.

| ordering | zone key, per cell | tests, ambiguous | random | on-land | answers moved | ambiguous query |
|---|---|---:|---:|---:|---:|---:|
| shipped | vertices ↑ (w = 0) | — | — | — | — | — |
| tie-break | vertices ↑, area ↓ breaking ties | +0.0 % | +0.0 % | −0.1 % | **0** | not timed |
| w = 0.25 | vertices / area^0.25 ↑ | −0.6 % | −0.4 % | −1.2 % | **0** | **+4.2 %** |
| w = 0.5 | vertices / area^0.5 ↑ | −1.2 % | −0.5 % | −1.6 % | 1 | **+5.5 %** |
| w = 1 | vertices / area ↑ — Smith's rule | −1.7 % | −2.0 % | −2.4 % | 2 | **+9.7 %** |
| w = ∞ | area ↓ — likeliest zone first | −2.8 % | −3.6 % | −3.6 % | 158 | **+24.8 %** |
| area ↑ | likeliest zone last, in the free slot | +1.3 % | +0.2 % | +1.1 % | 153 | not timed |
| w = ∞, block-capped | area ↓, cost capped at four blocks | −2.7 % | −3.5 % | −3.5 % | 158 | not timed |

**The family is monotone in the area weight and its optimum is `w = 0`, the shipped key.** More area weight buys fewer tests, moves more answers, grows the index (610,210 to 624,849 bytes, because distinct orderings deduplicate less well than distinct candidate sets) and costs more time, with no turning point between. So what is refused is the family, not one key.

Every timed row is slower on ambiguous with both estimators agreeing. `w = ∞` also reads +25.0 / +8.0 / +9.4 % on numba; `w = 1` was run twice (+10.8 % and +9.7 %, 8 and 4 of 61 rounds won) because the control drifted on the first. The small blends are *no difference* on `random` and `on-land` — the loss is confined to ambiguous, which is ~40 % of a mixed wall clock. The `unique` control drifted slightly *toward* the challenger on three runs (37–47 of 61, best round within 2 %, so `unresolved` rather than a claim), so the ambiguous losses are if anything understated. The three untimed rows have no side that could win.

**Why the count pointed the wrong way, which is the durable part.** A count of tests removed is a workload share only when the things counted cost the same, and here they systematically do not: the largest-overlap zone is usually the one with the most vertices, so any probability-led key makes the *first* test the expensive one. Weighting each cell's candidates by overlap area over the 31,368 ambiguous cells gives **+101 % expected vertices tested** against −2.8 % tests. The A/B confirms that direction and not that size — the query loses far less than the vertices predict, because the latitude block index makes a test sublinear in the ring. Sublinear, not free, which is the result.

Smith's rule (`w = 1`) is what the cost model says to do about that, and it does help: it keeps a huge polygon in the free final slot, which is why it moves 2 answers where `w = ∞` moves 158. It still loses, and so does every weight between, because today's key already minimises the term that dominates — the cost of a *failed* test. Sweeping the weight is what turns a refusal of one key into a refusal of a family, and it is the cheap half: the counts are one pass over the fixtures per key.

**This is a different instrument from the 2.90 % the entry was ranked on, not the same number re-run.** That was an index-uniform enumeration at resolution 3 over 41,162 cells; these are fixture-weighted counts at resolution 4. The near-match is a coincidence of denominators, not corroboration.

## The optimum, constructed and verified

The sweep above refuses a family by sampling it. Constructing the optimum closes the subject.

Under the loop's own cost model — `E = Σ_{t<S} c_t · (1 − Σ_{u<t} p_u)`, with `S` the index where the final zone's run starts — the ordering that minimises expected compute is: within a zone by `c/p` ascending; across zones by `C/P` ascending, since a zone's contribution at survival `R` is `R·C − K` and the `K` terms cancel under adjacent-block exchange, so a zone behaves **exactly** as one job of summed cost and summed probability; then enumerate which zone takes the free final slot, because designating `m` last both saves `R_m·C_m − K_m` and costs `P_m·Σ_{j>m}C_j` in lost shielding. **Verified against exhaustive enumeration of every contiguity-respecting ordering on all 31,344 cells with at most 5 candidates: optimal in 31,344 of 31,344.**

**The cost coefficient is not the vertex count.** The packed kernel skips latitude blocks, so `c_i = α + β·blocks + γ·in-band edges`, every count computable at build time from `_block_index.block_latitude_ranges`. Calibrated against measured `PolygonArray.pip` times over 1,500 (polygon, latitude) samples: **`437.5 ns + 0.511·blocks + 1.133·in-band edges`, R² = 0.9886**, against **R² = 0.524 for the total vertex count** the shipped key sorts on. A polygon of ≥512 vertices scans a median 2.2 % of its ring, and the two keys disagree on which candidate is cheaper in 31.6 % of pairs. The shipped proxy is genuinely stale.

**And it still does not matter, because the loop is overhead-bound.**

| | loop compute per ambiguous query | |
|---|---:|---|
| shipped | 758.5 ns | — |
| true-cost key | 735.8 ns | −2.99 % |
| **optimal** | **723.8 ns** | **−4.57 %** |
| oracle | 151.6 ns | −80 % |

437.5 ns per candidate is fixed and ~1.05 candidates are tested, so the whole loop is **14.7 % of a 5,150 ns ambiguous query**. The optimum saves 34.7 ns — **0.67 % of an ambiguous query, 0.27 % of a mixed workload**, against a 3–9 % noise floor. Three paired runs agree: +3.8 %, −6.3 %, +0.2 % on ambiguous with 29, 30 and 30 of 61 rounds won, `unresolved` or `no difference` throughout. Both the true-cost key and the optimum move **zero answers**, unlike every area-led key.

**Dropping zone contiguity for everything but the final zone** — the only place it is load-bearing, since `last_zone_change_idx` needs a suffix and `_zone_id_among` returns the zone of whichever polygon hits — is correct and buys nothing. Under it the prefix is unconstrained and the ratio rule applies per polygon; that construction is verified optimal by enumeration too (0 mismatches, 31,354 cells) and is worth **0.04 ns per ambiguous query, −0.005 %**. Contiguity *could* bind on 2,020 of 31,368 cells and actually costs something on **16**, worth −5.77 % on those alone. It binds only where a prefix zone contributes two or more polygons and another zone exists to interleave with, which the two-polygon majority rules out.

Even an oracle that knows the answer saves 4.7 % of a mixed workload. **That is the ceiling on candidate ordering as a subject**, and it is recorded as such: the loop is bound by per-call overhead, not by geometry, so future work here should attack the overhead rather than the order.

## What else the measurement turned up

- **Re-ordering candidates moves answers, with no data change** — 2 to 158 of 5,000 ambiguous points depending on the key. Every moved point lies inside **two** zones' overlapping upstream polygons, mostly `Asia/Urumqi` inside `Asia/Shanghai`, so which zone answers is decided by whichever is tested first. Nothing runs the data-update guard for an index reordering, so a future key owes that diff by hand. `tzfpy` agrees with the reordered answer on almost all of them — evidence about `tzfpy`'s precedence, not about correctness; nothing here derives a correctness property from the shortcut index.
- **Resolution 4 shrank the prize**: with `last_zone_change_idx` making the final zone free, 29,372 of 31,394 ambiguous cells hold exactly two polygons of two zones, where every order costs exactly one test. At resolution 3 it was 9,046 of 10,511.
- **The index grows** 2.0–2.4 %, because distinct orderings deduplicate less well than distinct candidate sets.
- Rebuilding the shipped ordering from source reproduces the packaged binary to **29 bytes**, across the few cells where the sort key ties and `polys_in_cell`'s set iteration order breaks the tie. Recorded as TOOL-7 in #648 rather than fixed here.

## The experiment is now committed and re-runnable

`prototypes/shortcut_ordering_optimum.py` holds the whole study as cached stages — `cells`, `areas`, `costs`, `calibrate`, `enumerate`, `build`, `count`, `ab` — so re-checking the refusal against a new dataset, H3 resolution or kernel is one command, and adding an ordering is one function in its `STRATEGIES` table. The module docstring carries the model, the cost and probability derivations, the construction with its proof, and **every assumption and approximation enumerated A1–A24**, each stating whether its error is measured or not.

Two assumptions that were previously asserted are now measured by the script:

- **Polygon overlap** (A14, the load-bearing one — (1)'s survival `1 − Σp` is exact only for disjoint hit regions). Candidates overlap in **1,090 of 31,368 cells (3.47 %)**; the doubly-covered area is 2.7 % of an affected cell on average and **100 % of the worst**, a fully nested pair. Modelling it properly is not a small correction — survival becomes the prefix's *union*, the ratio rule stops being a sort key, and the problem becomes correlated pipelined-filter ordering, which is hard in general. So the script takes the disjoint model and then measures every ordering it produces, with `count` reporting the answers each moves.
- **Single-zone coverage**: 826 cells (2.63 %) are covered entirely by one zone. That became PERF-7, in #648.

The A/B was also re-taken against the index the committed script builds — calibrated rather than guessed constants — and now reads **no difference on every stratum with both estimators agreeing** (+0.1 % ambiguous, 30 of 61 rounds; control clean), where the earlier guessed-constant runs read `unresolved` and swung ±6 %. Calibrating is what made the comparison quiet, which is itself a reason to run `calibrate` before believing an ordering.

## The two items it found go elsewhere

Both discoveries made beside this work — **TOOL-7** (the shortcut binary is reproducible only by set-iteration luck) and **PERF-7** (826 cells are covered entirely by one zone and could skip geometry entirely, worth 2.12 % of an ambiguous query, but 824 of them change which of two overlapping zones answers) — are in [#648](https://github.com/jannikmi/timezonefinder/pull/648), branched from `master` rather than stacked here, so this pull request retires exactly the item it ships. The prototype's `FINDINGS` still records both measurements, since the script is what produced them.

## Behaviour impact

None. Nothing that ships changed.

## Verification

The prototype was run end to end and reproduces every number in this body. Re-gated after rebasing onto current `master` (#646 reordered the ranking) and after removing the two items: **2,027 passed**. `uv run pre-commit run --all-files` (all hooks pass; `make hook` cannot install over the shared `core.hooksPath` in a worktree). `make testall`'s extra coverage is slow dataset and generator sweeps, neither of which this diff reaches.

Measured against `c27b452` on Apple arm64, free-threaded CPython 3.14.2, data 2026c, fixture set v3, both acceleration backends.

## Review

One independent review before opening, handed the diff and the routed rules but not the item or the reasoning. Findings taken: the bare 2.8 % relabelled as an ambiguous-stratum share in all three summary sites; the claim that the A/B "agrees with" the +101 % vertex figure corrected to agreement in direction only, with the block index named as why the sizes differ; the resolution-3 enumeration and the resolution-4 fixture count separated instead of presented as one number re-run; both directions of the key reported on the same strata; the `last_zone_change_idx` premise restored under the two-polygon figure; the counting rule's "any one of those clears the bar" conclusion qualified rather than left standing beside its own corrected premise; and the measurements de-duplicated — the item file holds them, the decision entry points there. Also flagged and acted on: the missing base SHA, coordinate-access mode, which on-land set, and the denominator of the vertex figure.

## Judgment calls

- Split the resolution record out rather than trimming existing entries for room. The budget rule says split, and renaming the parent file would have touched `timezonefinder/shortcut_index.py` and `timezonefinder/utils.py` for no reader's benefit.
- Kept a `Closed` entry rather than deleting it under the redundant-route exception. This idea is upstream's own suggestion and has already been re-proposed once after being refused; the entry is also where the four keys' numbers live, which is stated so a later pass does not delete it as redundant.
- Corrected the counting rule in place rather than noting the exception only in the item. The rule as written is what would send the next pass down this road.

## Verification of the analytic work

The construction is checked by exhaustive enumeration rather than argued; the cost model is checked by regression against measured kernel times rather than assumed. A derivation panel was also launched to attack the construction independently, and it died on a usage limit without returning — so the construction rests on the enumeration and the calibration, both reproducible from the recipe in the item.

## Next eligible item

GH-362 — reuse the `PolygonArray` binaries in file conversion — is the highest-ranked free row.


jannikmi:
Superseded by #650, which carries the shortcut-ordering work forward with geometrically integrated full-predicate costs, overlap-aware survival, and the final-zone-only ordering contract. The investigation and measurements from this PR informed that implementation, so this draft no longer needs to merge separately.

PR #650, opened 2026-09-08
Optimize shortcut ordering for full-predicate work during data conversion

jannikmi:
Compile ambiguous-cell shortcut orders from geometrically integrated full-predicate work. No query points are sampled, and no candidate-count cutoff excludes complex cells. The tested prefix may interleave zones; one complete final zone remains untested. Dropping zone contiguity primarily simplifies the optimization constraint.

At the maintainer's request this replaces the sampled objective and incorporates #653's enclosing-hole-box improvement. The branch is updated against merged #653 (`87626f5`), retaining its final `HoleArray` ownership and vectorized union-bound construction. The optimizer models its gate: individual hole work is charged only inside both the outer polygon box and the box enclosing its holes. The investigation in #638 motivates accounting for bounding-box rejection, holes and packed PIP blocks instead of vertex count alone.

The derivation, formulas and assumptions live beside the implementation in `scripts/shortcut_ordering.py`:

- Surface probability uses spherical area, with density `cos(latitude)` in longitude/latitude coordinates. Stored straight coordinate-plane edges have analytic area integrals; H3's spherical edges are deterministically subdivided and checked against its cell area. Cells are not assumed equal-area. The spherical-versus-WGS84 measure bound, floating overlay error, query quantization and the limits of the cell-area convergence check are documented.
- A positive rectangular-indicator cost model accounts for outer boxes, the enclosing hole box, individual hole boxes, PIP dispatch and latitude-active packed blocks. Rounded engineering coefficients are explicit; conversion never times code. Charging all holes and outer PIP even after a successful hole match is a deliberate approximation. Cache effects, comparison short circuits and backend differences are unmodeled.
- For already-tested set `S`, incremental work is the integral of the next check's point-dependent cost over `cell minus union(hit regions in S)`. This preserves dependence between reach probability and check cost. Overlaps use a union, not summed probabilities.
- Enumerate final zones, construct greedy orders by marginal hit area per integrated cost, and refine with adjacent exchanges. Also refine each legacy-derived prefix. Retain the legacy incumbent unless modeled work improves. This is a heuristic for the general cost model; the constant-cost, disjoint-hit special case gives the exact decreasing probability/cost ordering. There is no claim of globally optimal real latency, exponential subset enumeration, or a candidate-count fallback.
- An independent geometric safety gate preserves legacy zone precedence and the final zone where coverage/disjointness cannot be certified, while still optimizing within tested zone blocks. Invalid geometry is repaired only for the objective and always fails this gate. All candidates remain. Shared-border conventions remain a documented limitation.

`TimezoneFinderL` uses the new fallback-zone heuristic as explicitly approved; it need not pick the largest-area zone. No new binary layout or runtime ordering metadata is needed. PERF-7 from #648 remains separate by maintainer direction: no covered-cell collapse or new overlap precedence. No prototype code is included.

The regenerated binary changes 8,221 cell orders and retains every candidate in all 288,122 cells. All five cells above 14 candidates are evaluated; the 31- and 16-candidate cells change order. Only `shortcuts.bin` changes relative to the original dataset: 617,065 bytes, SHA-256 `fa36aec27a6f8d46a7b3270c5356d2d1524f11cd52ddd366502db7dd46af3f0e`. Generated binaries are not committed.

Validation: all 3,967 tests pass with `make testall` on Python 3.13.11. All 38 optimizer/hole-union tests also pass in an isolated default test environment. Tests cover analytic area at poles, the date line and all resolution-4 pentagons; exhaustive special-case optimality; point-dependent and overlapping costs; tiny positive hit regions; 51-candidate lists; and preserved zone answers for every synthetic hit pattern. Data integrity, candidate membership, the frozen 10,000-point answer guard (zero changes), generated documentation and all hooks pass.

No additional local benchmarks were run, as requested. Earlier CI timing results describe preceding revisions; CI must assess this implementation with its replacement data artifact.

The [CI-built data wheel](https://github.com/jannikmi/timezonefinder/actions/runs/34282827169) matches every local data file byte-for-byte. `DATA_BUILD_RUN` records this successful producer, so the subsequent PR matrix and benchmarks consume this dataset.


scripts/shortcut_ordering.py
```diff
@@ -0,0 +1,290 @@
...
+  Constants 1/N are omitted in code without changing the minimizer.
+* For constant c_i and disjoint hits, exchanging adjacent a,b changes the cost
+  by p_b*c_a-p_a*c_b: sort ALL nonterminal polygons by decreasing p_i/c_i,
+  enumerate the terminal zone, and choose the least J. Probabilities are still
+  unconditional within the cell, not renormalized after excluding that zone.
+  Our point-dependent model needs the recurrence instead of a ratio sort.
+* The optimum is exact for the sampled work objective, NOT measured latency or
+  the population integral. 128 samples are a fixed conversion-time accuracy
+  budget, with no claim that small slivers are resolved. States cost O(N n 2^n);
+  above 14 candidates the legacy order is retained, not called optimal.
```
jannikmi: these are the most important cells!

PR #652, opened 2026-09-08
PERF-7: answer a fully covered ambiguous cell without geometry

jannikmi:
## What changed

`scripts/zone_precedence.py` is new: it decides which of two overlapping timezone zones takes precedence, and turns an ambiguous shortcut cell that **one boundary polygon covers entirely** into a unique-zone entry. No point in such a cell can fall outside the covering polygon, so the whole point-in-polygon loop for it disappears; the reader already understands a zone-id entry, so nothing in `timezonefinder/` changes behaviour.

What made it non-trivial is that a covered cell *always* holds candidates of other zones — it would already be a unique-zone entry otherwise — and every one of them overlaps the coverer inside that cell. So the overlap has to be ordered before the cell can be collapsed.

**The precedence rule** is the one recorded on the item and decided 2026-09-08: of two overlapping zones, the one for which the shared area is the larger share of *itself* is the enclave and wins. It is derived from the zones' geometry rather than configured, computed once per zone pair, and applied only inside a covered cell.

## The item said this needs `shapely`. It does not, and a pass may not add it

The [improvement-pass workflow](contributing/workflows/run-one-improvement-pass.md) forbids changing dependencies or the lockfile, and `shapely` appears nowhere in `pyproject.toml`. Both halves turn out to be computable with what is already here:

- **Coverage** is `fully_contained_in_hole(cell, polygon)` — already in `scripts/utils_numba.py`, already exactly "closed ring A lies entirely inside closed ring B" — plus a check that no hole of the covering polygon meets the cell. `Hex.common_frame` is extracted so `lies_in_cell` and the new `covers_cell` cannot drift apart on the antimeridian frame they share.
- **Precedence** needs no intersection at all. The two ratios the rule compares share the shared area as their numerator, so `shared/area(A) > shared/area(B)` holds exactly when `area(A) < area(B)`. Areas are measured on the sphere (`spherical_ring_area`), not by a shoelace sum over the stored plate-carrée coordinates, so a high-latitude zone is not ordered ahead of a tropical one by the projection.

That reduction also explains an observation the item recorded without explaining it: the refused "smaller total zone area wins" alternative disagreed with containment on **0 of 118 pairs** because, inside the scope where it is applied, it is not a proxy for the rule — it *is* the rule. The decision record now says so, and keeps the two alternatives that are genuinely different (lower zone id; deciding per cell) as refused.

## Measurement

Compiled the full 2026c dataset twice — this branch and its merge base — from the same source GeoJSON, and diffed:

| | before | after |
|---|---:|---:|
| zone entries | 256,728 | 257,470 |
| candidate-list entries | 31,394 | 30,652 |
| `shortcuts.bin` | 610,210 B | 610,210 B |

`diff -rq` over the two compiled directories lists **`shortcuts.bin` and nothing else** — every other binary is byte-identical, and the index does not change size.

Point-in-polygon tests over the committed fixtures, counted by wrapping `inside_of_polygon` (a **count**, machine-independent, which is what the [ranking rules](contributing/improvements/improvement-ranking-and-eligibility.md) prefer over a clock):

| stratum | points | pip before | pip after | delta | answers moved |
|---|---:|---:|---:|---:|---:|
| ambiguous | 5,000 | 5,263 | 5,125 | **−2.6 %** | 2 |
| unique | 5,000 | 0 | 0 | — | 0 |
| random | 10,000 | 1,115 | 1,092 | **−2.1 %** | 0 |
| on_land | 10,000 | 1,998 | 1,910 | **−4.4 %** | 0 |

The two moved answers are `Africa/Khartoum → Africa/Juba` — the enclave the rule now names, at (9.50266, 28.27734) and (9.37275, 28.65404). This is a count over one sample, not a property: the rule *decides* precedence, so it changes answers by design wherever it differs from the ordering today happens to produce.

**This is not a suite-visible speed-up and is not claimed as one.** 2.12 % of an ambiguous query sits under the batch suite's 3–9 % noise floor; the removed-test count is the evidence, per the sub-floor rule.

## One recorded number did not reproduce

The item recorded "743 of 826 cells convert, **0 unresolved**". Re-derived here: **826 covered cells, 742 convert, 84 stay ambiguous.** The 84 are cells a larger zone covers while a *smaller* zone reaches into them without covering them — Xinjiang's boundary, mostly. Precedence gives the smaller zone the part it reaches, so part of the cell answers one way and part the other and no single zone can answer the whole cell. Converting them would be a wrong answer, so the sound count is 742/84, not 743/0.

## The rule is not the one `prototypes/hole_precedence_relation.py` refuted

That study finds size-derived precedence failing on 20 of 216 edges, systematically on the ocean zones — `Etc/GMT-3` must precede `Africa/Djibouti` there while being 583× its area, the opposite of what this rule says. The two never meet: those edges exist to answer a point inside a *hole* its own zone no longer claims, and `covers_cell` refuses coverage outright when any hole reaches the cell. Stated in the module docstring so nobody reads the two as contradicting.

## How it reaches users

The packaged data is not committed and `DATA_VERSION` stays `2026c`, so this changes no shipped byte today: it takes effect with the next compiled dataset. No `DATA_FORMAT_VERSION` or layout-version bump — the layout is untouched, only which cells qualify for an entry the reader already handles.

## Behaviour impact

None until the data is recompiled. After that: those cells answer without geometry; `TimezoneFinderL.timezone_at` and `unique_timezone_at` answer 742 cells they previously returned `None` for; and in a genuine two-zone overlap inside such a cell the enclave wins by rule rather than by whichever candidate the ordering tested first.

## Verification

- `uv run pre-commit run --all-files` — all hooks pass (`make hook` cannot install over the shared `core.hooksPath` in a worktree)
- `make test` — 2,050 passed (baseline 2,029 + the 21 added)
- `make testall` — **3,955 passed**, on this head against current `origin/master`
- `make testparse` — converter runs clean on the fixture input
- `make docs` — builds with no new warnings
- Two full converter runs over the real 2026c source, plus the `diff -rq` and fixture measurements above
- New tests pass with numba **genuinely unimportable** (`using_numba = False`, the pure-Python kernels), not merely with `NUMBA_DISABLE_JIT`

## Independent review

Taken before opening, on `8a62956`; five findings, four fixed and folded into the implementation commit, and the final head is ``039d414``.

- `unique_timezone_at` documented `None` for "multiple zones in this shortcut" — falsified by this change. Fixed.
- `shortcut_index.py`'s authoritative table-semantics docstring said a zone id means one zone covers the cell. Fixed.
- `docs/benchmarking_methodology.rst` tied "reads no geometry" to a single-zone cell. Fixed.
- `zone_areas` / `resolve_covered_cells` took an unannotated `data`, leaving a new seam unchecked by mypy. Fixed.
- **Declined, and recorded instead:** `covers_cell` proves containment of the cell's stored *planar* ring, while h3's real edges are geodesics that bow away from it by up to a few tens of metres at resolution 4. `lies_in_cell` already rests on the same approximation and a missed overlap there is a wrong timezone too, so this adds no new class of error — but the error direction is no longer benign, so it is stated in the docstring rather than left for a reader to discover.

## Judgment calls

- **A word-budget split was owed here and is paid.** `geometry-data-format-and-validation-decisions.md` sat at 1,999 of its 2,000-word limit, so this is the change that pushes it over. The shortcut index's binary shape and its six refused alternatives move out **verbatim** into `shortcut-index-binary-shape-decisions.md` (diffed against the original to prove nothing was dropped), leaving 1,686 and 643.
- The decision itself is recorded in `decisions/`, not in the deleted item: the rule and its mechanics live at `scripts/zone_precedence.py`, and the refused alternatives in the decisions file where a future pass looks before re-proposing one.
- Ties are refused rather than broken: two covering zones of equal area fail the strict `<`, so the cell keeps its candidate list and no set-iteration order reaches the binary (which is TOOL-7's subject, still open and untouched here).

## Ranking

PERF-7 was third on the ranking; `GH-362` and `GH-301` sit above it. It was taken because **the maintainer named it explicitly** for this pass, not because it outranked them — `GH-301` also has an open draft pull request (#638) and would have been yielded to regardless. Both remain unclaimed and eligible.

PERF-7 was recorded by #648, which had to merge before this could start: the claim protocol forbids basing a pass on an open pull request to absorb its contributor-memory edits, and this branch is off the merged `master`. The claim ref `refs/heads/improvement-claims/PERF-7` is released with this pull request.

## Next eligible item

`GH-362` — reuse the `PolygonArray` binaries in file conversion.


jannikmi:
Review finding (blocking correctness): `Hex.covers_cell()` establishes containment only for the planar ring joining H3 boundary vertices (`scripts/hex_utils.py:355-382`), then `resolve_covered_cells()` promotes that result to an unconditional zone answer (`scripts/zone_precedence.py:130-138`). Actual H3 cell edges curve outside that ring by tens of metres, as the new cell-local geometry record on `master` documents. A boundary can therefore contain every stored chord while excluding a valid point in the real H3 cell; after compilation that point silently receives the covering zone without any point-in-polygon check. The existing `lies_in_cell()` approximation does not make this safe: its conservative failure retains extra candidates, while this failure removes the query-time correctness check. Please replace coverage with a certified conservative enclosure/exact H3-cell predicate, and remeasure which cells still convert, before this optimization writes unique-zone entries.

jannikmi:
Deferred pending #656 / `GEOM-1`. Review found that the planar ring through `h3.cell_to_boundary` vertices is not a conservative model of every point H3 assigns to the cell. That affects the existing candidate-generation exclusions and blocks this PR’s stronger full-cell claim.

Resume this PR only after `GEOM-1` establishes a shared conservative geometry authority for both overlap and full containment, then rebase, replace `covers_cell` with that predicate, regenerate the index, and remeasure the cells that still convert. Marking this PR draft in the meantime.

jannikmi:
Superseded as an implementation by #691. Two things changed under this PR: #686 adds `timezones_at`, which reads exactly the candidates this conversion throws away (every converted cell is a two-zone overlap); and `master` has since blocked PERF-7 on GEOM-3, whose question is the planar-ring coverage claim this PR's own review declined. #691 records the route that keeps the candidates (7 extra entries on 2026c, but a format-4 release) and carries over this PR's measurements (742/84, no shapely). Leaving this as a draft for its converter code and measurements; not to be rebased or merged.

jannikmi:
Closing: PERF-7 is being withdrawn. A critical review found that applying containment precedence only inside cells one polygon covers makes one overlap region answer differently per H3 cell: in the Juba/Khartoum overlap (Abyei), 217 sampled points in converted cells would answer Juba while 322 in neighbouring cells keep Khartoum. That is the per-cell outcome the precedence decision refused, and it contradicts the rule that answers never depend on the H3 index. Elsewhere the conversion changes no answer, and what remains is ~0.6 % / 1.9 % / 2.1 % of a random / on-land / ambiguous query, against a certified-coverage proof and the candidates `timezones_at` needs. The withdrawal, with this branch's measurements kept as evidence, and a new item for applying the precedence rule through the candidate ordering in every cell, follow in a new PR. The commits stay reachable from this PR.

PR #683, opened 2026-09-15
Add a batch break-even and saturation benchmark

jannikmi:
## Summary

The batch API amortises validation, coordinate scaling and the shortcut table read over a whole batch, but still pays one `h3` cell lookup per point. So a small batch is *slower* per point than a loop of scalar calls, and a large one stops improving — and nothing in the repository measured either boundary. `benchmarks/test_timezone_finding.py` exercises the batch API at the single fixed `BATCH_SIZE` of 2,500, which is one point on the curve in question.

A caller sizing their batches now has an answer, and can measure their own: both boundaries move with the workload as much as with the machine, because points sharing an H3 cell are answered together, so the script takes `--points your.csv`.

**Measured on this branch** (clang, memory-mapped, 61 rounds × ~25,000 points per rung): both strata break even between N=20 and N=50, and reach **1.60x** (uniformly random) / **1.85x** (unique-shortcut). The ceiling independently matches the `~1.6x` recorded in `contributing/improvements/query-performance-measurement-baseline.md`, which was derived by a different instrument.

## Changes

- `scripts/measure_batch_break_even.py` — sweeps a log-spaced ladder of batch sizes, one `compare_candidates` call per rung, and derives the crossing interval, the saturation rung and a control statistic. Takes `--points` (a `lng,lat` CSV), `--api {names,ids}`, `--sizes`, `--rounds`, `--points-per-round`.
- `scripts/batch_break_even_chart.py` — the sweep as two panels on a shared log x axis: best-round speed-up above, round-level win share below.
- `docs/benchmark_results_batch_break_even.rst` + `docs/batch_break_even_sweep.svg` — the generated page and chart, wired into `make reports`, the `render-reports` CI job and `benchmark_report_artifact.REPORT_FILES`.
- `make batch-break-even`, and a new `benchmark` dependency group holding seaborn/matplotlib/pandas.
- Docs: toctree entry, `7_performance.rst` and `1_usage.rst` pointers, a methodology section on sweeping one parameter.
- Contributor memory: the candidate-comparison rules move to their own module (`designing-a-candidate-comparison.md`), which is what kept `benchmarking-and-performance-validation.md` inside its word budget.

### Design points that are not obvious

Each is recorded at its decision site and in contributor memory:

- **Work per round is held constant across rungs**, and per-point figures divide by what a round actually answered rather than by the nominal target. A fixed call count per round would put the smallest and largest rungs in different measurement regimes and plot them as one curve.
- **Batches are drawn without replacement.** A repeated coordinate is answered from the batch's own cell lookup and the saving grows with N — it would fake exactly the amortisation being measured. The ladder is capped at half the pool for the same reason: at N equal to the pool there is one possible batch, and drawing it every round is not sampling.
- **The ladder brackets `NAMES_GATHER_MIN_BATCH`, imported rather than hard-coded.** `ZoneNames.names_of` switches from a Python loop to a numpy gather there, so a plain 1-2-5 ladder steps over a real discontinuity inside the measured path and renders it as noise between 100 and 200.
- **`compare_candidates` charges ~0.21 µs per timed item**, which per-point normalisation spreads as `h/N`: 16.5 % of the scalar baseline at N=1, under 1 % by N=50. It does not move a crossing above N≈20, but it ramps the smallest rungs of a curve that should be flat, so the control statistic excludes them rather than flagging them every run.
- **A crossing is read off the within-rung ratio; a saturation point off the batched side's absolute time.** Between N=1,000 and N=2,000 the ratio moved 2.9 % while the batched per-point time moved 0.2 % — all of it scalar-baseline noise.
- **The answer is published as an interval, never a single size.** A discrete ladder plus a threshold verdict cannot locate a crossing, and the rungs inside the interval read `unresolved` by construction — which the chart draws as hollow markers rather than leaving to read as a broken run.

### Why two panels rather than a distribution plot

A box/boxen plot over rounds was considered and rejected: it would show the spread of per-round timings, which `docs/benchmarking_methodology.rst` defines as measurement noise rather than a difference in what was measured, and which is why the tracked estimator is `min`. The two panels instead draw the harness's two estimators, so "believe a difference only where both agree" is shown rather than asserted — the win share crosses 0.5 inside the same band where the speed-up crosses 1.0.

### Packaging

seaborn moves out of `proto` into a new `benchmark` group, which `proto` pulls back with `include-group`. The group is deliberately outside `test` and outside `$(BENCHMARK_ENV)` so no measured process imports the plotting stack — `make memory` records `memory::import::rss` in subprocesses of that environment. matplotlib carries a version **ceiling** as well as a floor, because a minor upgrade rewrites the committed SVG wholesale.

## Related issue

None.

## Validation

- [x] `uv run pytest tests/ -q` — 4084 passed, 1 skipped
- [x] Other relevant checks:
  - `pre-commit run --all-files` — all hooks pass (mypy, rstcheck, check-manifest, ruff)
  - `sphinx-build -W --keep-going` — builds clean, the SVG is copied
  - **Determinism**: page and chart re-rendered from the same JSON in separate processes are byte-identical. Without `svg.hashsalt`, `svg.fonttype="none"` and `metadata={"Date": None}` the SVG differs on every render; `tests/test_batch_break_even.py` holds this.
  - **Isolation**: `uv run --isolated --group test --group compare python -c "import importlib.util as u; assert u.find_spec('matplotlib') is None"` passes, so the plotting stack is absent from the measurement environment.
  - **Stability**: three independent `make batch-break-even` runs compared; see the comment below for the result.

## Checklist

### Known limitations

- The committed figures are one machine's (an Intel Xeon @ 2.10 GHz, not the CI runner class). The page says so, and says the answer moves with the caller's own clustering.
- `control_spread` is a range over a dozen rungs, so its 15 % threshold is much looser than the 3 % a paired comparison resolves. It catches gross contamination and nothing finer; that is what a range statistic can honestly do, and it is published rather than used as a gate.
- `--api ids` is offered but not published: `timezone_ids_at` has no public scalar counterpart, so it is a bound rather than a like-for-like pair.

https://claude.ai/code/session_01NyjwmYqZ2877fAWYJ65Xtm

---
_Generated by [Claude Code](https://claude.ai/code/session_01NyjwmYqZ2877fAWYJ65Xtm)_


jannikmi:
## Review round: five findings, all fixed in `0671725`

Four came from an independent review at medium effort; the fifth from the run-to-run stability check this change's own plan required. Each was reproduced before being fixed.

### 1. All ~25 derivation tests skipped in the environment CI actually uses

`pytest.importorskip("seaborn")` sat at **module level**, which skips the whole file rather than the six chart tests. `build.yml` syncs `--group test`, and `benchmark` is a separate group:

```
$ uv run --isolated --group test pytest tests/test_batch_break_even.py -q
1 skipped in 0.04s          # before
34 passed, 8 skipped        # after
```

The skip is now taken inside each chart test.

### 2. `break_even` called a table full of `faster` rows "never faster"

The terminal-run search returned `not_reached` whenever the *last* rung alone failed to read `faster`, and then reported every genuinely faster rung as non-monotone under a note referring to an interval that was never printed. Reproduced with seven clearly-faster rungs and one unresolved top rung:

```
verdicts    : slower, faster x7, unresolved
status      : not_reached
describe    : "not demonstrably faster at any rung measured"
non_monotone: (5, 10, 20, 50, 100, 200, 500)
```

An empty terminal run now splits in two: `not_reached` when nothing won, and a new `no_terminal_run` when something did but the ladder does not end in a win — which says exactly that and asks for a re-measurement.

### 3. The renderer ignored the thresholds the run was taken under

`build_report` stamps `saturation_tolerance`, `control_spread_threshold` and `control_spread_min_batch_size` into the JSON, and nothing read them back — `grep` found all three only in the writer. Re-rendering a stored run after a constant moved would silently restate answers that run never made. They are now read back, with the constants as the fallback for older runs; the rule `get_batch_size` already enforces for `BATCH_SIZE` on the other pages.

### 4. `load_points_csv` rejected a header that follows a comment

The header skip keyed off `line_number == 1`, so a `--points` file opening with `# ...` rejected its own header. It now skips on the first row that is neither blank nor a comment.

---

### 5. Saturation does not reproduce on a drifting machine — the page was publishing through it

The plan's gate was that the break-even bracket and the saturation rung must agree across three runs. Three independent `make batch-break-even` runs:

| run | random: saturation | random: control | unique-shortcut: saturation | unique-shortcut: control |
|---|---|---|---|---|
| 1 (committed) | **N=1,000** | 5.0 % ok | **N=500** | 7.5 % ok |
| 2 | **N=500** | 4.3 % ok | **N=500** | 18.5 % **above** |
| 3 | **not reached** | 15.2 % **above** | **not reached** | 21.8 % **above** |

**The crossing held at N=20–50 in all six sweeps.** That is the primary question, and it reproduces.

**Saturation did not**, and the control says why. The asymmetry is structural rather than bad luck: a crossing is a paired comparison *within* one rung, so drift over the run cancels out of it; saturation is the batched call's absolute per-point time compared *across* rungs, which drift moves directly. That trade-off was deliberate — taking saturation from the ratio instead would mix in scalar-baseline noise, measured at 2.9 % against 0.2 % — but the page published the number regardless of whether the run could support it.

The instrument was working: it detected its own bad conditions and was ignored. So the fix is to gate the claim, not to patch the number. The headline now states the saturation figure only where the control certifies the ladder measured one thing, and says the control failed otherwise. This is also the behaviour a downstream user wants on a noisy laptop, which is the point of shipping the script.

The committed run passes on both sweeps (5.0 % and 7.5 %), so its published figures are unchanged. More rounds were deliberately *not* added as a fix: drift is not what `min` over rounds removes, and raising `--points-per-round` would cost every user runtime on an unproven theory.

### Validation

- `uv run --group benchmark pytest tests/ -q` — **4089 passed, 1 skipped** (5 new tests)
- `uv run --isolated --group test pytest tests/test_batch_break_even.py -q` — 34 passed, 8 skipped
- `pre-commit run --all-files` — all hooks pass
- `sphinx-build -W --keep-going` — clean
- Page and chart re-rendered from the same JSON in separate processes: byte-identical
- Measurement environment still carries no matplotlib

### Known gap, deliberately left open

The six chart tests — including the SVG-determinism guard — still skip in CI. `tox.ini` runs `uv run pytest`, which resolves uv's default groups, so making them run means adding `--group benchmark` to a tox env rather than to the workflow's `uv sync`. That is a wider change than this round warranted, and the matplotlib version ceiling in `pyproject.toml` is the primary defence against the failure it guards. Worth a follow-up.

---

jannikmi:
## Follow-up closed here, and the branch is current

Two things since the last round, in `2bdfb45` and `7ce3b4f`.

### The chart tests now run in CI

The previous comment left this open as "wider than an obvious fix". Reading `tox.ini` made that wrong — it is one standalone env plus one matrix string, and the same pattern had just landed on master for `free-threaded` (#680). So it is closed here rather than deferred.

`[testenv:charts]` runs `tests/test_batch_break_even.py` under `--group benchmark`, and `charts` joins the 3.12 matrix entry:

```
$ uv run tox -e charts
42 passed in 8.54s          # was 34 passed, 8 skipped in the plain environment
```

The SVG-determinism guard — the one thing between a matplotlib upgrade and a permanently churning committed chart — therefore now runs somewhere other than a contributor's laptop.

Kept out of `envlist`, like `[testenv:benchmarks]`: a bare local `tox` should not pull matplotlib and pandas into a fresh environment, and `make install` already gives a contributor every group. `tests/test_python_version_support.py` needed no edit — it reads standalone envs off `tox.ini`'s sections rather than a hard-coded list, precisely so adding one costs nothing there.

The separation the group exists for still holds:

```
$ uv run --isolated --group test --group compare python -c \
    "import importlib.util as u; assert u.find_spec('matplotlib') is None"
clean
```

### Merged master (was 5 commits behind)

`2bfe4e4` and the four before it — the free-threading trio, the data-versions docs change and a cibuildwheel bump. Clean merge; `docs/1_usage.rst` was the only file both sides touched and the two edits are in different sections.

**The committed page and chart are deliberately not re-rendered.** Nothing measured changed: the merge's diff inside `timezonefinder/` is entirely docstrings and comments, with no executable line touched, and `NAMES_GATHER_MIN_BATCH` is still 128, so the ladder still brackets it and the page's figures still describe this tree. Re-rendering would only rewrite committed figures from a different machine than the one that produced them.

### Validation on the merged head

- `uv run --group benchmark pytest tests/ -q` — **4094 passed, 2 skipped** (master's free-threading test accounts for the change from 4089/1)
- `uv run tox -e charts` — 42 passed
- `uv run --isolated --group test pytest tests/test_batch_break_even.py -q` — 34 passed, 8 skipped
- `uv run pytest tests/test_python_version_support.py -q` — 9 passed (the tox/matrix guard)
- `pre-commit run --all-files`, `sphinx-build -W --keep-going`, `uv lock --check` — all clean
- `git status docs/` — the generated page and chart untouched

No open findings remain on this PR.

---

PR #689, opened 2026-09-20
Name the missing App credential instead of failing at the token step

jannikmi:
## Summary

The weekly data-update job for upstream `2026d` ([run #35484919324](https://github.com/jannikmi/timezonefinder/actions/runs/35484919324)) failed 2 seconds into `open_update_pr`, in `actions/create-github-app-token`:

```
Error: The 'client-id' (or deprecated 'app-id') input must be set to a non-empty
string. If using a secret or variable, ensure it is available in this workflow context.
```

`DATA_UPDATER_APP_ID` and `DATA_UPDATER_PRIVATE_KEY` are not set on this repository. **That was the first time the regeneration job had ever run** — every earlier weekly run exited at `update_needed=false`, and `release_data_update.yml`'s `merge_and_release` job has been skipped on all 826 of its firings — so nothing had ever asked for the credential and nothing could report that it was missing.

Two things then went wrong beyond the missing secret itself. The action's error names neither secret nor this repository, and it fires before anything in the job can say what happened. And the fallback opened [#688](https://github.com/jannikmi/timezonefinder/issues/688) asking for the data to be regenerated by hand with `update_data.sh` — hours of converter, candidate audit, answer guard, fixture and report work — for a pipeline that is working and a cause that two secrets and a re-run fix.

This does not update the data and does not set the secrets; it makes the next failure say what it is.

## Changes

- `check_data_updates.yml`: the regeneration job checks both secrets before minting the token and fails with a message naming them, what to install, and why the default `GITHUB_TOKEN` is not a fallback (a branch it pushes triggers no workflow, so the update PR would carry data `build.yml` never validated and `release_data_update.yml` could never merge).
- `release_data_update.yml`: the same check, for a sharper reason — it is triggered by `workflow_run`, so its failures appear on no pull request and in no issue, and an absent credential fails it *before* `resolve-update-pr` has found the pull request every notice in that job is keyed to. Naming the cause in the log is all that is left.
- Both checks read `${{ secrets.X != '' }}` rather than the values, so no credential reaches a shell where `set -x`, a crash or an error message could echo it.
- The fallback issue now distinguishes its two causes. It is the only place a scheduled failure is ever seen, and "the pipeline broke, compile the data by hand" and "this repository has no credentials to open a pull request with" ask for entirely different work.
- Three invariants in `tests/test_data_update_workflow.py`, which now covers both halves of the pipeline: every job minting an App token checks the credentials first, no step takes a credential as an environment *value*, and the fallback issue can tell its two causes apart. All three fail against the workflows as they stood.
- The data-pipeline memory gains one clause naming the two secrets as a prerequisite of both workflows.
- `Internal:` changelog fragment.
- **DOC-11**, riding along at the maintainer's request: a register item for sweeping out the historical illustrations that carry no rule, plus the regression guard. Raised by the review on this branch — the first draft of that memory clause grew four sentences of provenance. The item is the criterion rather than a site list (keep the incident when a competent contributor would otherwise reasonably break the rule; cut it when the rule stands alone and the story is provenance `git log` already holds), it is deletions only, and its guard is a rule rather than a check, because the only mechanical signal is a date and decision records are required to carry one.

## Related issue

Related to [#688](https://github.com/jannikmi/timezonefinder/issues/688) — this is the CI half of it. The `2026d` data itself is still packaged as `2026c` and is not regenerated here; that goes through the pipeline once the App is configured, not by hand.

Note that #688 is open, and the weekly job skips the notification when an issue with that title already exists — so the corrected message will not appear on it. I have left a comment there with the real cause.

## Validation

- [x] `make test` — 2144 passed, 1 skipped, 1915 deselected (2:01)
- [x] Other relevant checks:
  - `pytest tests/test_data_update_workflow.py` — 14 passed; the three new tests fail (3 failed, 11 passed) with the workflow changes stashed, so they pin the fix rather than the wording.
  - `pytest tests/test_improvement_ledger.py tests/test_contributor_memory.py` — 11 passed, for the DOC-11 entry and its ranking row.
  - Both preflight scripts executed directly under `bash` with `APP_CONFIGURED=true` (exit 0) and `=false` (exit 1, `::error::` naming both secrets).
  - The fallback issue's script executed with a stubbed `gh` under both values, rendering the configured and unconfigured bodies.
  - `pre-commit run --files <changed>` — all hooks pass.
  - `yaml.safe_load` on both workflows.

These were run in the Claude Code container on this branch, not on a GitHub runner. What is *not* validated here is the workflows actually executing: a scheduled run cannot be exercised from a pull request, and the credential path needs secrets this branch cannot see. The tests read YAML, which is what the existing invariants in that file do and why its docstring limits them to failures that are otherwise silent.

## Checklist

## What is still blocked on you

The automation cannot run until a GitHub App with write access to contents and pull requests is installed on this repository and its App ID and private key are stored as `DATA_UPDATER_APP_ID` and `DATA_UPDATER_PRIVATE_KEY`. Once they are set, re-running `check_data_updates.yml` prepares the `2026d` release the way it was designed to.

https://claude.ai/code/session_011v1d2CBB1tUjXYjKvsU7fk


contributing/development/data-pipeline-format-versioning-and-release-order.md
```diff
@@ -19,5 +19,5 @@
...
-- `.github/workflows/check_data_updates.yml` opens an update PR weekly when upstream has a new release; `release_data_update.yml` merges it and pushes the tag with a GitHub App token (the default `GITHUB_TOKEN` would not trigger `build.yml`)
+- `.github/workflows/check_data_updates.yml` opens an update PR weekly when upstream has a new release; `release_data_update.yml` merges it and pushes the tag with a GitHub App token (the default `GITHUB_TOKEN` would not trigger `build.yml`). **Both read `DATA_UPDATER_APP_ID` and `DATA_UPDATER_PRIVATE_KEY`, and neither can do anything without them**, so each checks them before minting the token rather than meeting the action's own input error. Nothing exercised that credential until upstream published `2026d` on 2026-09-20: the weekly check had run for months, but always exited at `update_needed=false`, so the first release that needed the App is the one that discovered there was none. When the weekly job falls back to a notification issue, the issue says which of the two happened - an unconfigured App, which two secrets and a re-run fix, or a regeneration that genuinely failed, which is the one that needs `update_data.sh` run by hand
```
jannikmi: This addition is too verbose

PR #691, opened 2026-09-21
PERF-7: keep the candidates timezones_at reads, and price what that costs

jannikmi:
## What changed

This PR only edits the register. PERF-7's entry now records how the item keeps the candidates `timezones_at` reads, and what keeping them costs. The item stays **blocked on GEOM-3**, and no code or data changed.

## Why

`timezones_at` (API-1, shipped in #686) walks an ambiguous cell's whole candidate list and reports every zone containing the point. In a cell one polygon covers entirely, every other-zone candidate lies inside that polygon, so every point it reaches is in two zones. Each of the cells PERF-7 converts is therefore one where `timezones_at` needs both zones. Collapsing the cell to a zone id, which is what #652 does, would make it answer one. Converting only the cells whose other candidates are spurious keeps nothing worth having, so the conversion has to keep the list.

**The route recorded:** give a covered cell its own entry rather than a zone id. Entries are deduplicated on *(candidate list, covering zone)* and placed in a reserved index range, with a per-entry column holding the covering zone. `timezone_at`'s ambiguous branch returns that zone without geometry, `timezones_at` walks the list as it does today, and the unique-zone path is untouched.

**Priced on 2026c,** by recompiling with #652's converter:

| | entries |
|---|---:|
| today (distinct candidate lists) | 2,994 |
| with retention | 3,001 |

The 742 converted cells share only **seven** (list, zone) pairs, and the payload does not grow. Two costs remain:

- One integer comparison on every ambiguous query. It is unmeasured; an in-query A/B against the ~109 ns the conversion saves on average would settle it.
- **`SHORTCUT_LAYOUT_VERSION` 3 and `DATA_FORMAT_VERSION` 4.** `timezonefinder-data` 3.2026.3 has published, so this item would pay for an ordered two-distribution release alone. That is now its dominant cost, and the entry says it belongs with the next format change rather than opening one.

## Corrections carried over from #652's measurement

- **742 convert, 84 stay ambiguous — not the 743/0 the entry recorded.** The 84 are cells a larger zone covers while a smaller zone reaches in without covering them. Converting one of them would be a wrong answer.
- **No `shapely`.** Coverage is `fully_contained_in_hole`, already in `scripts/utils_numba.py`. Precedence reduces to comparing spherical areas.
- **"Smaller total zone area wins" was listed as a refused alternative, but within this scope it is containment itself.** The two ratios containment compares share the shared area as their numerator, which is why they disagreed on 0 of 118 pairs. GH-513's ocean-zone failures belong to the hole relation, a different relation.
- The entry cited `prototypes/shortcut_ordering_optimum.py`, which never landed. It now points at draft #652 as the evidence.

## #652

It stays a **draft**. It discards the candidates `timezones_at` reads, and it implements an item that `master` has since blocked on GEOM-3. Its planar-ring coverage claim is the one that item's audit found unsound, and that its own review had declined to fix. It is kept as evidence rather than rebased.

## Overlap with #686 (merged first)

#686 added a bullet to this same entry naming three candidate routes. Resolved on this branch as agreed: the bullet is replaced by the *Keeping the candidates* section, which carries its one requirement the section lacked, a test that reads coverage independently of the index. References to `timezones_at` and `tests/test_overlapping_zones.py` now point at what shipped rather than at the open PR. The sequencing line #686 added stays.

## Verification

- `tests/test_improvement_ledger.py` and `tests/test_contributor_memory.py` pass (11 passed); entry at 1,615 of its 2,000 words.
- `uv run pre-commit run --all-files` is clean. `make hook` can't install over the shared `core.hooksPath` in a worktree, so the hooks were run directly.
- No changelog fragment: the contributor-memory layer is exempt.

## Next eligible item

This PR doesn't change it. On the current ranking that is `TOOL-7`, as #686 also reports.


jannikmi:
Independent review, run after opening (it should have come first), on `0099077`: one finding. The ranking row's Size cell still said M while the entry now says M-L; fixed in `bc9b8b2` (amended), register tests pass. No other findings.

PR #700, opened 2026-09-21
Re-measure the headline comparison against pytzwhere

jannikmi:
## What

- **New prototype** `prototypes/pytzwhere_headline_comparison.py` measures this package against its predecessor `pytzwhere` (3.0.3, the last release) on one sample of points spread evenly over the globe. It reports startup time, peak and added memory, cold and per-query latency (whole globe and land only), coverage, per-query crashes, on-disk size, and whether `pytzwhere` still starts on a current Python/NumPy stack.
- **`docs/alternatives.rst`**: the *Comparison to pytzwhere* section quoted only the fork-era "76 MB CSV / 450 MB RAM" story. It now also covers the last release and the re-measured magnitudes.
- `prototypes/README.md` row and a `documentation/` changelog fragment. Under the changelog policy the prototype itself needs no fragment.

## Why

The origin story is the main reason this package exists, but nothing in the repository measured it again. The origin story also predates `pytzwhere`'s last release, which swapped the CSV for gzipped GeoJSON.

## How it measures

- Each package runs in its own subprocess, so neither one's imports or memory leak into the other's numbers. Each query is timed individually after one warm-up query; the cold first query is reported separately.
- `pytzwhere` runs in an isolated `uv run --isolated` environment pinned to Python 3.10, `numpy<1.24` and `shapely<2`, the newest stack it still works on. The unpinned attempt is run as well, and its error is part of the report.
- `pytzwhere` raises `KeyError` for latitudes missing from its shortcut table (near the poles). Those are recorded per query as crashes instead of aborting the run.

## Results (third of three 20,000-point runs, timezonefinder-data 3.2026.4, macOS arm64)

This package is measured once per point-in-polygon path. `clang` (the C extension) is what a plain `pip install` gets; `numba` needs the optional extra, and `python` is the fallback where no extension is built. All three paths gave identical answers.

| metric | tzwhere | tzf numba | tzf clang (default) | tzf python |
|---|---|---|---|---|
| startup (import + init) | 1.50 s | 0.25 s | 0.07 s | 0.06 s |
| RSS added by import + init | 597 MiB | 92 MiB | 18 MiB | 17 MiB |
| median latency, global / land | 7.5 / 11.0 µs | 1.0 / 1.0 µs | 0.9 / 0.9 µs | 1.0 / 1.0 µs |
| p99 latency, land | 794 µs | 9.7 µs | 6.7 µs | 1,474 µs |
| answered, global / land | 28.9 % / 84.2 % | 100 % | 100 % | 100 % |
| queries that raised | 48 | 0 | 0 | 0 |
| installed data size | 22.9 MiB | 32.1 MiB | 32.1 MiB | 32.1 MiB |

- **numba carries no headline and costs ~75 MiB.** The default install is the strongest column.
- **The pure-Python fallback's p99 is slower than pytzwhere's**, whose ray cast runs in shapely's C code. The docs page says so.

## For the reviewer

- **The docs state magnitudes only.** The documentation rules forbid copying exact measured figures into hand-written prose, so the page gives magnitudes and names the script. The exact run lives in the script's `FINDINGS` block.
- **p99 is quoted only qualitatively.** The clang and python tails were stable across runs. The numba tail was not: it put 111–165 µs outliers into single runs, so no numba p99 is stated.
- **On-disk size is the one regression**, and the page says so.
- **Same-zone agreement on land (~82 %)** compares `tz_world` with a current release, so it measures a decade of border and zone-name changes as much as either implementation.
- **Data version:** measured on the 2026d release. An earlier numba-only run on 2026c moved every non-p99 row by under 20 %.


jannikmi:
## The rendered report pages are waiting in an artifact

`3d9e87c` adds a section to the acceleration-path page (*What installing Numba costs before any query*), so the committed pages need the render that goes with it. It ran and succeeded against this exact head: **[benchmark.yml run 35671854097](https://github.com/jannikmi/timezonefinder/actions/runs/35671854097)** (`render_reports=true`, artifact `benchmark-pages`, 25 KB, expires 2026-12-21).

I could not install it from this session: GitHub redirects artifact downloads to `productionresultssa8.blob.core.windows.net`, which this session's egress policy denies (`connect_rejected`). That is a transport restriction on my side only — the artifact itself is fine.

To install it:

```bash
rm -rf tmp/benchmark-pages
gh run download 35671854097 --name benchmark-pages --dir tmp/benchmark-pages
uv run python -m scripts.benchmark_report_artifact install \
  --artifact-dir tmp/benchmark-pages \
  --expected-commit 3d9e87c9940f72d286ecb0b883fe0d53f6e7b0e5
```

The stamp check refuses the artifact against any other tree, so it only applies while this head stands; a later push needs its own dispatch. Commit the pages on their own, staging `docs/benchmark_results_*.rst` and `docs/data_report.rst`.

Everything else on this head is green: `build`, the PR benchmark run (all rows unchanged), and Read the Docs.

---

PR #701, opened 2026-09-22
PERF-9: bind the C extension wherever it loaded, numba only where it did not

jannikmi:
## Summary

The point-in-polygon dispatch in `timezonefinder/utils.py` preferred Numba whenever it was importable. That ordering predates both the C extension and the packed payload the kernels read today, and the measurement runs the other way: the JIT kernel is slower than the extension on every polygon size and never faster on a whole lookup. This binds the extension wherever it loaded and leaves Numba as the fallback for an installation that has none — which is the case the extra was always needed for, and where its rival is the pure-Python fallback rather than the extension.

An installation holding both now gets the faster kernel without reinstalling anything. It does **not** get its memory back: `utils.py` imports `utils_numba` either way, so a process with Numba installed still compiles that module's helpers and still holds the compiler. The docs say so where they offer the extra.

**Measured on this branch** — Intel(R) Xeon(R) @ 2.80GHz, Python 3.12.3, fixture version 3, `scripts/measure_acceleration_paths.py`, 61 rounds × 2,500 inputs per row, order alternated inside one process:

| row | numba relative to clang | rounds won | verdict |
|---|---|---|---|
| packed kernel, small polygons | 1.73x | 0 of 61 | slower |
| packed kernel, medium polygons | 1.78x | 0 of 61 | slower |
| packed kernel, large polygons | 1.93x | 0 of 61 | slower |
| `timezone_at`, random points | 1.20x | 4 of 61 | slower |
| `timezone_at`, unique-shortcut points | 1.00x | 40 of 61 | unresolved |
| `timezone_at`, ambiguous-shortcut points | 1.25x | 0 of 61 | slower |

The unresolved row is the shortcut index answering outright, with no point-in-polygon test to move — the one workload no kernel choice affects. These are local numbers, not for any committed page; the committed report is regenerated on a runner as usual.

**The side traded away**, measured the same way (median of three cold processes, `scripts/_startup_probe.py`): a numba-installed process holds **152 MiB** resident once it has answered, against **48 MiB** for a plain install. Before this change the same process held 186 MiB, so binding the extension recovers ~34 MiB of it and no more. Nothing regresses — the plain install is untouched — but the extra now costs that for nothing wherever the extension built.

## Changes

- `timezonefinder/utils.py`: `if clang_extension_loaded:` replaces `if clang_extension_loaded and not using_numba:`, with the reasoning and the reopening condition recorded at the decision site.
- `scripts/assert_acceleration_path.py`: new `--expect-interpreted`. With the extension outranking Numba, "the environment with numba installed" is no longer expressible as an *active* path, and the acceleration-path comparison — which rebinds the kernels itself — needs to assert exactly that.
- **The guard that inverting the dispatch silently weakened** (found in review, `78bfe40`): `--expect clang` used to imply "and no numba", because clang could only be active where numba was absent. Every measurement target and both `benchmark.yml` assertion steps now assert the interpreted path as well, with a per-target regression test verified to fail without the fix. Without it, a developer machine — where `make install` always brings numba — could have put a ~100 MiB-inflated `memory::import::rss` and JIT-compiled kernel rows into the committed pages and the trend chart.
- **Report labelling, which the change would otherwise have broken silently:** `acceleration_path_label` and `describe_benchmark_machine.acceleration_label` read the numba flag first, so a clang run in a numba-holding environment would have been published as "Numba JIT". Both now read the binding (`using_clang_pip`) first.
- Page prose on `benchmark_results_acceleration_paths`: the dispatch description, and a closing paragraph that claimed installing Numba swaps `validate_coordinates` for two `njit`'d helpers — stale since #644 removed those decorators; it now describes what installing Numba actually still changes.
- Docs (README, getting started, performance, architecture, benchmarking methodology) and the contributor memory that stated the old ordering, including the three `prototypes/` docstrings whose invocation recipes were labelled by the path they used to select.
- `TimezoneFinder.using_numba()` is unchanged in meaning and unchanged in value; `using_clang_pip()` keeps reporting the binding. On an installation with both, both are now `True`.

## Related issue

None — this came out of the acceleration-path measurements added in #700.

## Validation

- [x] `make testall` — **4186 passed, 1 skipped**; `make test` **2258 passed, 1 skipped** after the review fixes
- [x] `make hook` — clean
- [x] `make docs` — builds with no new warnings
- [x] Unit suite in the **numba** environment (`uv run --isolated --group test --group numba pytest -m "not integration and not slow"`) — 2223 passed, 25 skipped. This is the configuration the change alters, so it is gated separately.
- [x] `scripts/measure_acceleration_paths.py` in the numba environment, both before and after the dispatch change — the table above.
- [x] Both assertion modes exercised by hand in both environments, including the failure the guard now produces in a numba environment.
- [x] The CI fix in `019b4e6` verified by running its shell in both trees: this branch (has the make target) yields `--expect clang --expect-interpreted python`, a master checkout yields `--expect clang`.

New tests: `test_the_extension_outranks_numba_wherever_it_loaded` covers all four environments the dispatch decides between by re-executing `utils` under both flags, since any one session can only exercise its own branch; `test_every_measurement_target_also_refuses_a_numba_environment` and the workflow equivalent pin the guard; `test_headline_section_warns_on_a_clang_run_that_also_holds_numba` pins the reporting case this change creates.

## Checklist

- [x] Documentation and the changelog were updated when appropriate (`changelog.d/changed/clang-extension-outranks-numba.rst`).
- [x] Packaging, binary data, or configuration changes are explained above — none; packaged data is untouched.

### Risks and limitations

- **One machine, one dataset.** The ranking is measured here and on the CI-rendered page, both pointing the same way, and #700's macOS run agreed. It is not measured on every architecture, and the decision record names what would reopen it: a measurement putting numba ahead on the packed kernel.
- **The `py*-numba` tox environments now bind clang**, so the JIT kernel is no longer covered incidentally by their whole-suite runs. It is covered explicitly — `tests/test_acceleration_paths.py` drives the full lookup stack on each path by binding it — which is stronger than the incidental coverage it replaces, but it is a deliberate change in how that path is reached.
- **The committed report pages still describe the old dispatch** until they are re-rendered on a runner; the renderer prose in this branch is what they will pick up.
- **A trap worth knowing about**, which cost this branch one red run: `benchmark.yml`'s merge-base step runs *this* workflow against the merge base's code, so any make target or CLI flag a pull request adds must be optional there. `019b4e6` makes the new flag conditional for exactly that reason.
- Process note: this ran as an improvement pass under the claim `improvement-claims/PERF-9`, but no register item file was created — the item would have been added and retired in the same pull request, leaving no net register change. The durable reasoning is at the decision site instead, per the core contract. **The claim ref could not be deleted** from the session that made it (ref deletion fails there, while ordinary pushes succeed), so `git push origin :refs/heads/improvement-claims/PERF-9` is still owed by hand — otherwise a later pass reads it as a live claim.

https://claude.ai/code/session_01TNADsQis61F4gZzc3cB3PT


jannikmi:
## Review round: one real defect, fixed in `78bfe40`

The independent review found that this change silently weakened a guard, and it is the kind that fails quietly.

`--expect clang` used to carry a second meaning for free. With numba winning the dispatch whenever it was importable, clang could only be *active* where numba was absent — so asserting the active path also asserted the environment. Inverting the dispatch took that away: `--expect clang` now passes with numba installed. Only `acceleration-paths` had been given the replacement assertion, so `benchmarks`, `latency`, `memory`, `batch-break-even`, `benchmarks-ci`, `memory-ci` and both assertion steps in `benchmark.yml` kept an assertion that no longer refuses what it was written to refuse.

What that would have admitted is not a mislabelled kernel — the kernel is the same one now — but a process *holding* numba: `memory` records its resident set, which the import inflates by ~100 MiB, and `benchmarks` times `test_pt_in_poly_python*` under node ids the trend chart joins on, which numba JIT-compiles. On a developer machine that is the default state, since `make install` syncs every group.

Fixed by asserting the interpreted path beside the active one (`BENCHMARK_INTERPRETED_PATH := python`) at every measurement site, plus `print-benchmark-interpreted-path` for the two workflow steps.

The test that existed only checked that the script was invoked at all, which is exactly why the weakening passed it. It now checks the flag, per target and across the workflow — verified failing without this commit (6 failed) and passing with it, and the assertion verified by hand to refuse a numba environment:

```
ERROR: expected this environment to hold the 'python' kernel, but it holds 'numba' (utils.using_numba=True).
```

Two documentation findings from the same round: the methodology page claimed a stray numba install "still has to fail the assertion" — true only once this landed, and it now names the mechanism; and two prototype docstrings said "both backends must be measured" above two recipes that now both bind clang, where the distinction is the environment rather than the backend.

Gate on this head: `make test` 2258 passed / 1 skipped, `make hook` clean, `make docs` no new warnings, affected test files 264 passed.

---

